PR #3, opened 2019-01-26
Add more object functions.

Frando:
Hi, started using napi-rs instead of neon. Works very well so far.

I needed a few more functions to work with js objects and arrays and added them here. Originally based on the fork from @cztomsik  but rebased onto this master here.


cztomsik:
wow, I had no idea anyone else could be beneficient from my additions, hope it helped you and thanks for pinging :-)

Brooooooklyn:
thanks！

Issue #55, opened 2020-05-25
Decouple sys module of new napi-rs into a crate


Brooooooklyn:
https://github.com/napi-rs/napi/issues/4

PR #77, opened 2020-06-19
feat(napi): impl threadsafe function api

oyyd:
~This is a premature implementation for getting more comments.~

### DIFF
--- a/napi/Cargo.toml
+++ b/napi/Cargo.toml
@@ -20,3 +20,12 @@ glob = "0.3"
 napi-build = { version = "0.1", path = "../build" }
 regex = "1.3"
 semver = "0.10"
+
+[features]
+default = []
+# See the N-API Version Matrix https://nodejs.org/api/n-api.html#n_api_n_api_version_matrix
+napi2 = []
+napi3 = []
+napi4 = []
+napi5 = []
+napi6 = []
--- a/napi/src/threadsafe_function.rs
+++ b/napi/src/threadsafe_function.rs
@@ -1,0 +1,233 @@
+use crate::{check_status, ptr, sys, Env, Function, Result, Value};
+use std::os::raw::{c_char, c_void};
+
+use sys::napi_threadsafe_function_call_mode;
+use sys::napi_threadsafe_function_release_mode;
+
+pub trait ToJs: Copy + Clone {
+  type Output;
+  type JsValue;
+
+  fn resolve(
+    &self,
+    env: &mut Env,
+    output: Self::Output,
+  ) -> Result<(u64, Value<Self::JsValue>)>;
+}
+
+/// Communicate with the addon's main thread by invoking a JavaScript function from other threads.
+///
+/// ## Example
+/// An example of using `ThreadsafeFunction`:
+///
+/// ```
+/// #[macro_use]
+/// extern crate napi_rs_derive;
+///
+/// use std::thread;
+/// use napi_rs::{
+///   Number, Result, Value, Env, CallContext, Undefined, Function,
+///   sys::{
+///     napi_threadsafe_function_call_mode::{
+///       napi_tsfn_blocking,
+///     },
+///     napi_threadsafe_function_release_mode::{
+///       napi_tsfn_release,
+///     }
+///   }
+/// };
+/// use napi_rs::threadsafe_function::{
+///   ToJs, ThreadsafeFunction,
+/// };
+///
+/// // Define a struct for handling the data passed from `ThreadsafeFunction::call`
+/// // and return the data to be used for the js callback.
+/// #[derive(Clone, Copy)]
+/// struct HandleNumber;
+///
+/// impl ToJs for HandleNumber {
+///   type Output = u8;
+///   type JsValue = Number;
+///
+///   fn resolve(&self, env: &mut Env, output: Self::Output) -> Result<(u64, Value<Self::JsValue>)> {
+///     let argv: u64 = 1;
+///     let value = env.create_uint32(output as u32)?;
+///     Ok((argv, value))
+///   }
+/// }
+///
+/// #[js_function(1)]
+/// fn test_threadsafe_function(ctx: CallContext) -> Result<Value<Undefined>> {
+///   // The callback function from js which will be called in `ThreadsafeFunction::call`.
+///   let func: Value<Function> = ctx.get::<Function>(0)?;
+///
+///   let to_js = HandleNumber;
+///   let tsfn = ThreadsafeFunction::create(ctx.env, func, to_js, 0)?;
+///
+///   thread::spawn(move || {
+///     let output: u8 = 42;
+///     // It's okay to call a threadsafe function multiple times.
+///     tsfn.call(Ok(output), napi_tsfn_blocking).unwrap();
+///     tsfn.call(Ok(output), napi_tsfn_blocking).unwrap();
+///     // We should call `ThreadsafeFunction::release` manually when we don't
+///     // need the instance anymore, or it will prevent Node.js from exiting
+///     // automatically and possiblely cause memory leaks.
+///     tsfn.release(napi_tsfn_release).unwrap();
+///   });
+///
+///   Ok(Env::get_undefined(ctx.env)?)
+/// }
+/// ```
+#[derive(Debug, Clone, Copy)]
+pub struct ThreadsafeFunction<T: ToJs> {
+  raw_value: sys::napi_threadsafe_function,
+  to_js: T,
+}
+
+unsafe impl<T: ToJs> Send for ThreadsafeFunction<T> {}
+unsafe impl<T: ToJs> Sync for ThreadsafeFunction<T> {}
+
+impl<T: ToJs> ThreadsafeFunction<T> {
+  /// See [napi_create_threadsafe_function](https://nodejs.org/api/n-api.html#n_api_napi_create_threadsafe_function)
+  /// for more information.
+  pub fn create(env: &Env, func: Value<Function>, to_js: T, max_queue_size: u64) -> Result<Self> {
+    let mut async_resource_name = ptr::null_mut();
+    let s = "napi_rs_threadsafe_function";
+    let status = unsafe {
+      sys::napi_create_string_utf8(
+        env.0,
+        s.as_ptr() as *const c_char,
+        s.len() as u64,
+        &mut async_resource_name,
+      )
+    };
+    check_status(status)?;
+
+    let initial_thread_count: u64 = 1;
+    let mut result = ptr::null_mut();
+    let tsfn = ThreadsafeFunction {
+      to_js,
+      raw_value: result,
+    };
+
+    let ptr = Box::into_raw(Box::from(tsfn)) as *mut _ as *mut c_void;
+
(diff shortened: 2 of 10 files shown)


test_module/__test__/tokio_readfile.spec.js
```diff
@@ -0,0 +1,20 @@
...
+
+const filepath = path.resolve(__dirname, './example.txt')
+
+test('should read a file and return its a buffer', async (t) => {
+  return new Promise((resolve, reject) => {
+    bindings.testTokioReadfile(filepath, (err, value) => {
+      try {
+        t.is(err, null)
+        t.is(Buffer.isBuffer(value), true)
+        t.is(value.toString(), 'Hello world!\n')
```
Brooooooklyn: Maybe `t.is(value.toString(), fs.readFileSync(filepath, 'utf8'))` could fix this case in windows.

napi/src/threadsafe_function.rs
```diff
@@ -0,0 +1,168 @@
...
+      sys::napi_call_threadsafe_function(
+        self.raw_value,
+        Box::into_raw(Box::from(value)) as *mut _ as *mut c_void,
+        mode,
+      )
+    })?)
+  }
+
+  pub fn acquire(&self) -> Result<()> {
+    Ok(check_status(unsafe {
```
Brooooooklyn: just
```rust
check_status(unsafe {
  sys::napi_acquire_threadsafe_function(self.raw_value)
})
```
no need Ok(xxx?)

oyyd: Yes. Will remove all the unnecessary `Ok(...)`.

napi/src/threadsafe_function.rs
```diff
@@ -0,0 +1,168 @@
+use crate::{check_status, ptr, sys, Env, Function, Result, Value};
+use std::os::raw::{c_char, c_void};
+
+// TODO Should we pub these enums? or pub the sys mod directly?
+pub use sys::napi_threadsafe_function_call_mode;
```
oyyd: @Brooooooklyn Do you think we should `pub sys` mod?

Brooooooklyn: I will publish `sys` as an independent crate `https://github.com/napi-rs/napi-rs/issues/55`

Brooooooklyn: BTW, We have already `pub mod sys;` in `lib.rs`

napi/Cargo.toml
```diff
@@ -19,4 +19,8 @@ bindgen = "0.54"
 glob = "0.3"
 napi-build = { version = "0.1", path = "../build" }
 regex = "1.3"
-semver = "0.10"
\ No newline at end of file
+semver = "0.10"
+
+[features]
+default = ["tsfn"]
```
oyyd: I have added a `tsfn` feature flag here because it's `N-API 4` and I think it can't get compiled for lower version of Node.js.

But it seems that we could compile the addon in a "higher" version of Node.js but use it on a lower version. (Still, we should not use higher N-API if we want the addon actually work for lower version of Node.js.) Therefore I will remove this `tsfn` flag.

napi/src/threadsafe_function.rs
oyyd: And then I would like to stop `pub` these enums.

napi/Cargo.toml
Brooooooklyn: Just few lines configuration should handle this situation properly. 

```rust
// napi/lib.rs
#[cfg(napi4)]
pub mod tsfn;
```

```rust
// napi/build.rs
let napi_version = Command::new("node")
  .args(&["-e", "console.log(process.config.variables.napi_build_version)"])
  .output()
  .unwrap()
  .stdout;
for version in [1..napi_version.parse().unwrap()] {
  println!("cargo:rustc-cfg=napi{}", version);
}
```

Brooooooklyn: seems like `node@10` doesn't have `napi_build_version` property in `process.config.variables`, maybe we should set `node@10.0.0 ~ node@10.16.0` version "3" and `node@10.16+` version "4" here according to the https://nodejs.org/api/n-api.html#n_api_n_api_version_matrix.

napi/src/threadsafe_function.rs
```diff
@@ -0,0 +1,157 @@
...
+      to_js,
+      raw_value: result,
+    })
+  }
+
+  pub fn call(
+    &self,
+    value: Result<T::Output>,
+    mode: napi_threadsafe_function_call_mode,
+  ) -> Result<()> {
```
Brooooooklyn, line 144: maybe return `()` here, and `napi_throw` if got error is better

test_module/src/lib.rs
```diff
@@ -112,3 +139,102 @@ fn get_external_count(ctx: CallContext) -> Result<Value<Number>> {
...
+struct HandleNumber;
+
+impl ToJs for HandleNumber {
+  type Output = u8;
+  type JsValue = Number;
+
+  fn resolve(&self, env: &mut Env, output: &mut Self::Output) -> Result<(u64, Value<Self::JsValue>)> {
+    let argv: u64 = 1;
+
+    let value = env.create_uint32(*output as u32).unwrap();
```
Brooooooklyn: change `unwrap()` to `?`

napi/src/threadsafe_function.rs
Brooooooklyn: so does in `release` function, developer who are using `tsfn` may call these two funtions in `Thread::spawn`, and they can do nothing if error happened.

napi/src/threadsafe_function.rs
```diff
@@ -0,0 +1,157 @@
...
+use sys::napi_threadsafe_function_release_mode;
+
+pub trait ToJs: Copy + Clone {
+  type Output;
+  type JsValue;
+
+  fn resolve(
+    &self,
+    env: &mut Env,
+    output: &mut Self::Output,
```
Brooooooklyn: why not just `Self::Output`?

oyyd:
Thanks, will check the comments later and supplement some docs.

napi/src/threadsafe_function.rs
oyyd: We don't use the `output` anymore so we can leave the owner of `output` to the user and avoid copying the `output`, can't we?

napi/src/threadsafe_function.rs
oyyd: > so does in release function, developer who are using tsfn may call these two funtions in Thread::spawn, and they can do nothing if error happened.

Yes. But I guess there is likely no other script on the stack so that `napi_throw` would cause an `uncaughtException`(actually, IDK, maybe we can't call any non-threadsafe api here because the V8 is paused here.)

test_module/src/lib.rs
oyyd: Fixed.

test_module/src/lib.rs
```diff
@@ -112,3 +139,102 @@ fn get_external_count(ctx: CallContext) -> Result<Value<Number>> {
...
+
+  fn resolve(&self, env: &mut Env, output: &mut Self::Output) -> Result<(u64, Value<Self::JsValue>)> {
+    // TODO avoid copying buffer
+    let value = env.create_buffer_with_data(output.to_vec())?;
+    Ok((1u64, value))
+  }
+}
+
+async fn read_file_content(filepath: &Path) -> Result<Vec<u8>> {
+  match tokio::fs::read(filepath).await {
```
Brooooooklyn: tokio::fs::read(filepath).await.map_err(|_| Error...)

napi/src/threadsafe_function.rs
```diff
@@ -0,0 +1,211 @@
...
+  to_js: T,
+}
+
+unsafe impl<T: ToJs> Send for ThreadsafeFunction<T> {}
+unsafe impl<T: ToJs> Sync for ThreadsafeFunction<T> {}
+
+impl<T: ToJs> ThreadsafeFunction<T> {
+  /// See [napi_create_threadsafe_function](https://nodejs.org/api/n-api.html#n_api_napi_create_threadsafe_function)
+  /// for more information.
+  pub fn create(env: Env, func: Value<Function>, to_js: T, max_queue_size: u64) -> Result<Self> {
```
Brooooooklyn: `env: &Env` is okay, so you needn't call `*ctx.env` in `js_function`

napi/src/threadsafe_function.rs
Brooooooklyn: Okay, I think call `expect` here would make scenes. I still think return `Result` which couldn't be handled outside isn't a graceful design.

```rust
check_status(unsafe {
  sys::napi_call_threadsafe_function(
    self.raw_value,
    Box::into_raw(Box::from(value)) as *mut _ as *mut c_void,
    mode,
  )
}).expect("call threadsafe function failed")
```

napi/src/threadsafe_function.rs
Brooooooklyn: > so we can leave the owner of output to the user 

Yes, but if you want give the ownership to the user in rust, you must `move` it, but not give the `borrowed` one to the user.

In most case, if the user want consume the `&mut Output`, they must call `*Output` or `Output.to_owned()` or something else, it's actually a `Copy` happened.
And in our case, `tsfn` and `async_callback`, we put the `Output` in the `heap memory` (Box) rather than `stack memory`. So pass `Box::from_raw(data)` to `resolve` function actually won't cause the `Copy` happened. It just `Copy` the smart pointer (Box) to the `resolve` under the hood.
The only problem is, we should use `output: Box<Self::Output>` to align the semantics. But I don't think this is a big problem here.

Brooooooklyn: > so we can leave the owner of output to the user 

Yes, but if you want give the ownership to the user in rust, you must `move` it, but not give the `borrowed` one to the user.

In most cases, if the user want consume the `&mut Output`, they must call `*Output` or `Output.to_owned()` or something else, it's actually a `Copy` happened.

If the use case of `tsfn` is really performance sensitive, the user can choose `Box` or `Cell` to represent the `Output` themselves.

napi/src/threadsafe_function.rs
oyyd: I have also changed the `refer` and `unref` to use a reference of `Env`.

napi/src/threadsafe_function.rs
oyyd: Change to `Self::Output`. 

I think we can add another api that won't `Copy`. My use case is creating a large `Buffer` of Node.js from `u8` without copying data and vice versa.

Brooooooklyn: > My use case is creating a large Buffer of Node.js from u8 without copying data and vice versa.

Yes, it seems to be a common use cases. Maybe we can create something like `NodeBuffer`, which mange the memory manually under the hood and could `deref/deref_mut` to `&[u8]/&mut [u8]`, and easy transform to `JsArrayBuffer/JsBuffer`

Brooooooklyn:
LGTM! let's squash and merge it

oyyd:
Thanks! Rebased.

PR #81, opened 2020-06-25
refactor(napi): redesign the JavaScript values API

Brooooooklyn:
# TODO
- [x] Update `README` and examples
- [x] Testing `node-rs`
- [x] More unit tests

# Background

The original `napi-rs` APIs was from very earlier `x-ray` project, and `x-ray` even dropped the `N-API` implementation before the `N-API` module got been stable.

So, I decide to redesign the APIs from old `x-ray`,  especially the `Value` struct and `ValueType` trait, it's too verbose for writing library based on `napi-rs`:

```rust
fn fibonacci(ctx: CallContext) -> Result<Value<Number>> {
  let n = ctx.get::<Number>(0)?.try_into()?;
  ...
}
```
vs
```rust
fn fibonacci(ctx: CallContext) -> Result<JsNumber> {
  let n = ctx.get::<JsNumber>(0)?.try_into()?;
  ...
}
```

The new `module` struct allow to export things easier:

```rust
exports.set_named_property("testThrow", env.create_function("testThrow", test_throw)?)?; 
```
 vs

```rust
module.create_named_method("testThrow", test_throw)?;
```

And the `Error` struct has been redesign either,  now the `napi-rs` could throw the right error messages from Rust errors.


oyyd:
Should this PR be a semver major change?

Brooooooklyn:
according to `semver`, versions below `1.0` don't need to follow `semver` semantic 🧠

Brooooooklyn:
@oyyd I will merge this pull request now, and some more jobs still need to be done before next version release, like https://github.com/napi-rs/napi-rs/issues/82 and https://github.com/napi-rs/napi-rs/issues/83 
And I will get those things done in next few days.

oyyd:
@Brooooooklyn Thanks for the reminder and sorry for not having enough time to help review the PR.

Personally, I would like to write a small lib to verify that `tsfn` could work well with tokio, or find out what is needed to be supplement in the progress.

PR #86, opened 2020-07-02
Use libuv to run rust future

Brooooooklyn:
- Add back `napi@2~3` support
- Testing on `node@10.0.0`
- Fix napi@version cfg in `napi-build`
- **Run future on `libuv`**


.github/workflows/napi3.yaml
```diff
@@ -57,7 +58,7 @@ jobs:
 
       - name: Unit tests
         run: |
-          yarn --ignore-engines
+          yarn add ava@2 --dev --ignore-engines
```
oyyd, line 61: is this necessary as `ava` has been added in the package.json?

napi/src/env.rs
```diff
@@ -416,6 +467,67 @@ impl Env {
...
+    resolver: R,
+  ) -> Result<JsObject> {
+    use futures::prelude::*;
+
+    let mut raw_promise = ptr::null_mut();
+    let mut raw_deferred = ptr::null_mut();
+
+    unsafe {
+      let status = sys::napi_create_promise(self.0, &mut raw_deferred, &mut raw_promise);
+      check_status(status)?;
```
oyyd, line 508: thought: Maybe we can add a function like `check_status` but allow appending a string to the `reason` field.

napi/src/promise.rs
```diff
@@ -0,0 +1,112 @@
...
+) {
+  let mut env = Env::from_raw(raw_env);
+  let future_promise = Box::from_raw(context as *mut FuturePromise<T, V>);
+  let value = ptr::read(data as *const _);
+  let js_value_to_resolve = (future_promise.resolver)(&mut env, value);
+  let deferred = future_promise.deferred;
+  let mut handle_scope = ptr::null_mut();
+  match js_value_to_resolve {
+    Ok(v) => {
+      let open_handle_status = sys::napi_open_handle_scope(raw_env, &mut handle_scope);
```
oyyd: `call_js_cb` is called in the main thread so that I guess napi should have opened a handle scope here and we don't have to call `sys::napi_open_handle_scope` again?

.github/workflows/napi3.yaml
Brooooooklyn: ava@3 is not compatible with node@10.0.0, so I have to downgrade it to 2.x here

napi/src/env.rs
Brooooooklyn: Good idea

PR #87, opened 2020-07-02
chore: add into_unknown; refactor tsfn api; add tests for js function

oyyd:
Just found that `ctx.get::<Function>(0)` is not implemented in the previous version. 

`JsFunction::call()` accept `JsUnknown` as its arguments but it seems that there is no simple way to create a `JsUnknown` from other js types, like `JsString` in the example. We might want to add a api for that?

### DIFF
--- a/napi/src/js_values/buffer.rs
+++ b/napi/src/js_values/buffer.rs
@@ -2,7 +2,7 @@ use std::ops::{Deref, DerefMut};
 use std::ptr;
 use std::slice;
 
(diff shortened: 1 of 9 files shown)


Brooooooklyn:
Maybe we could add `into_unknown` method in `impl_js_value_methods!` macro

oyyd:
Add `into_unknown` instead.

oyyd:
@Brooooooklyn Please take another look. I have also modified the api of `tsfn` and would like to merge this if it still looks good to you.

PR #89, opened 2020-07-06
Implement NativeBuffer to perform zero-copied manually drop  Buffer

Brooooooklyn:


Brooooooklyn:
`Box::from(Vec<u8>)` is already `zero copy`

PR #94, opened 2020-07-08
feat(napi): implement tokio_rt feature to execute future on tokio runtime

Brooooooklyn:


napi/src/env.rs
```diff
@@ -527,6 +534,36 @@ impl Env {
...
+  ) -> Result<JsObject> {
+    let mut raw_promise = ptr::null_mut();
+    let mut raw_deferred = ptr::null_mut();
+    check_status(unsafe { sys::napi_create_promise(self.0, &mut raw_deferred, &mut raw_promise) })?;
+
+    let raw_env = self.0;
+    let future_promise =
+      promise::FuturePromise::create(raw_env, raw_deferred, Box::from(resolver))?;
+    let future_to_resolve = promise::resolve_from_future(future_promise.start()?, fut);
+    thread::spawn(move || {
```
Brooooooklyn: spawn a system thread on every execute is really a bad idea

napi/src/tokio_rt.rs
```diff
@@ -0,0 +1,49 @@
...
+pub(crate) enum Message {
+  Task(Pin<Box<dyn Future<Output = ()> + Send>>),
+  Shutdown,
+}
+
+#[inline]
+pub(crate) fn get_tokio_sender() -> &'static mpsc::Sender<Message> {
+  static SENDER: OnceCell<mpsc::Sender<Message>> = OnceCell::new();
+  SENDER.get_or_init(|| {
+    let (sender, mut receiver) = mpsc::channel(100);
```
Brooooooklyn: Does buffer size `100` big enough? Maybe we should enable user config it via environment variable? 

What do you think about it @oyyd

Issue #92, opened 2020-07-07
Finish all N-API binding


Brooooooklyn:
API remains:

- [x] Enum types
- [x] napi_create_buffer_copy
- [x] napi_create_date
- [x] napi_create_typedarray
- [x] napi_create_dataview
- [x] napi_create_bigint_int64
- [x] napi_create_bigint_uint64
- [x] napi_create_bigint_words
- [x] napi_create_string_latin1
- [x] napi_get_prototype
- [x] napi_get_typedarray_info
- [ ] napi_get_dataview_info
- [x] napi_get_date_value
- [x] napi_get_value_bigint_int64
- [x] napi_get_value_bigint_uint64
- [x] napi_get_value_bigint_words
- [x] napi_get_value_string_latin1
- [x] napi_get_global
- [x] napi_instanceof
- [x] napi_is_error
- [x] napi_is_typedarray
- [x] napi_is_dataview
- [x] napi_strict_equals
- [x] napi_detach_arraybuffer
- [x] napi_is_detached_arraybuffer

oyyd:
@Brooooooklyn I would like to implement some apis. Maybe I could start with `bigint` related apis (e.g. `napi_get_value_bigint_int64`, `napi_get_value_bigint_uint64` and `napi_get_value_bigint_words`)?

Brooooooklyn:
@oyyd great, I would like finish https://github.com/napi-rs/napi-rs/pull/94 before anything else in https://github.com/napi-rs/napi-rs/milestone/1

PR #98, opened 2020-07-10
feat(napi): add BigInt related apis

oyyd:
Refs: #92


Brooooooklyn:
Should `u128` and `i128` also need to be implemented `TryFrom<JsBigInt>`?

And maybe `BigInt` and `BigUint` also need to be implemented `TryFrom<JsBigInt>` [num_bigint](https://docs.rs/num-bigint/0.3.0/num_bigint/) under feature flag.

Brooooooklyn:
Close in favor of https://github.com/napi-rs/napi-rs/pull/133

PR #108, opened 2020-07-26
Removed unshifted null from arguments of threadsafe callbacks

### LINKED ISSUE #107: Node.JS Async callback convention
Hello,

Here you prepend a null value to the arguments that you pass to the callback:

https://github.com/napi-rs/napi-rs/blob/2ff09ea26d0c2b7b5c520ff40db5ba1154e2e67e/napi/src/threadsafe_function.rs#L207-L213

I don't think this is a convention? I can't find any reference to this in the Node.JS documentation.

Would it be possible to remove this null and move that responsibility to the user if they need to have it?

I can make a PR if you prefer.

Mike-Dax:
Hello,

This PR removes the prepended arguments to threadsafe node callbacks. I've tested it with both arguments and a lack of arguments passed to the callback.

### DIFF
--- a/napi/src/threadsafe_function.rs
+++ b/napi/src/threadsafe_function.rs
@@ -202,21 +202,17 @@ unsafe extern "C" fn call_js_cb<T: ToJs>(
 
   let status;
 
-  // Follow the convention of Node.js async callback.
   if ret.is_ok() {
     let values = ret.unwrap();
-    let js_null = env.get_null().unwrap();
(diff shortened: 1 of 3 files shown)


Mike-Dax:
Upon reflection I think this is the wrong approach - I see that you're following this:

https://nodejs.org/en/knowledge/errors/what-are-the-error-conventions/

I'll make a PR to update the documentation to reflect that to make it clearer to people trying to use the library.

PR #181, opened 2020-09-19
refactor: build on Windows without admin privilege.

### LINKED ISSUE #180: Windows build: PermissionDenied
When building `napi-rs` on Windows, I get this error:
```
error: failed to run custom build command for `napi-package-template v0.1.0 (C:\Data\Projects\package-template)`

Caused by:
  process didn't exit successfully: `C:\Data\Projects\package-template\target\release\build\napi-package-template-7c08280ff4b54a5f\build-script-build` (exit code: 101)
  --- stderr
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 5, kind: PermissionDenied, message: "Access is denied." }', C:\Users\fluxx\.cargo\registry\src\github.com-1ecc6299db9ec823\napi-build-0.2.1\src\lib.rs:42:66
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Caused by this line of code:
https://github.com/napi-rs/napi-rs/blob/master/build/src/lib.rs#L42

It looks like the build script tried to write lib file to nodejs install folder which by default is 
```
C:\Program Files\nodejs
```
Is it better to save the lib file to the current user's home dir? Otherwise, we need Admin privilege to build napi-rs modules.

Thanks

fluxxu:
Fix #180 

Instead of trying to write to `C:\Program Files`, which requires admin privilege, this PR creates a `.napi-rs` folder inside the current user's home directory to save the downloaded `node-{version}.lib` file.

It seems that `napi-sys` crate downloads and extracts `node-headers` files to `OUT_DIR`:
https://github.com/napi-rs/napi-rs/blob/master/sys/build.rs#L61
Do you think we should move them to `.napi-rs` folder too? Looks like node-gyp is doing just that.
image
Thanks!

### DIFF
--- a/build/src/lib.rs
+++ b/build/src/lib.rs
@@ -9,29 +9,30 @@ use cfg_if::cfg_if;
 cfg_if! {
   if #[cfg(windows)] {
     use std::env::var;
-    use std::fs::File;
+    use std::fs::{File, create_dir};
     use std::io::copy;
     use std::path::PathBuf;
 
     pub fn setup() {
       let node_full_version =
         String::from_utf8(Command::new("node").arg("-v").output().unwrap().stdout).unwrap();
-      let node_exec_path = String::from_utf8(
-        Command::new("node")
-          .arg("-e")
-          .arg("console.log(process.execPath)")
-          .output()
-          .unwrap()
-          .stdout,
-      )
-      .unwrap();
-      let node_exec_path = node_exec_path.trim_end();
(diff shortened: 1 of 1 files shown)


Brooooooklyn (approved):

Brooooooklyn:
> Do you think we should move them to .napi-rs folder too? Looks like node-gyp is doing just that.

Yes, I think we should do it.

PR #270, opened 2020-10-31
Add Electron Support

### LINKED ISSUE #125: Electron Support
While experimenting with this library, I realized that I can't get any generated modules to be loaded by Electron on Windows. I'm running on Windows 10 and trying to use Electron 9.1.1. 

I think the problem is with linking to the wrong version of node.lib, as the build script appears to unconditionally link to the node.lib of the installed node, while it should be linking to electron's node.lib. I think this library is also missing a delay-load hook implementation.

adumbidiot:
This PR will add Electron support for Windows. I don't think any other platform has any significant hurdles, but I am unable to test this as I can only test on Windows.

Possibly closes #125.

### DIFF
--- a/build/Cargo.toml
+++ b/build/Cargo.toml
@@ -11,3 +11,6 @@ version = "0.2.1"
 
 [dependencies]
 cfg-if = "1.0"
+
+[target.'cfg(windows)'.dependencies]	
+ureq = { version = "1.5.0", default-features = false, features = [ "native-tls" ] }
--- a/build/src/lib.rs
+++ b/build/src/lib.rs
@@ -1,115 +1,18 @@
-extern crate cfg_if;
+cfg_if::cfg_if! {
+    if #[cfg(windows)] {
+      mod windows;
+      pub use windows::setup;
+    } else if #[cfg(target_os = "macos")] {
+      mod macos;
+      pub use macos::setup;
+    } else {
+      pub fn setup() {
+        setup_napi_feature();
+      }
+    }
+}
 
 use std::process::Command;
-
-use cfg_if::cfg_if;
-
-cfg_if! {
-  if #[cfg(windows)] {
-    use std::fs::{create_dir, metadata, write};
-    use std::path::PathBuf;
-    use std::io::Read;
-
-    fn get_node_version() -> std::io::Result<String> {
-        let output = Command::new("node").arg("-v").output()?;
-        let stdout_str = String::from_utf8_lossy(&output.stdout);
-
-        // version should not have a leading "v" or trailing whitespace
-        Ok(stdout_str.trim().trim_start_matches('v').to_string())
-    }
-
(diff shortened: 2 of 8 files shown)


adumbidiot:
I don't think that CI failure is related to the changes made. I'm not done yet, this is still WIP.

Job List
* It is missing a delay-load implementation.
* It doesn't respect npm environment variables in the build script.
* It uses a primitive js script to download node.lib, which breaks under simple circumstances such as trying to follow redirects for * electron's node.lib.
* The global cache for node.libs is sperated by version. However, Electron versions their node.lib by electron version, not nodejs version. This will lead to builds failing unpredictably if someone builds for Electron and node on the same computer.
* The linker args emitted by the build script appear to not be respected. I was able to find a solution only by moving the build script execution into the actual module's build script.
* The napi version feature is set by the executing nodejs, not the engine its targeting.

adumbidiot:
Are there any real benefits to using a global `node.lib` cache? It seems like a hassle to maintain correctly, it pollutes the home dir, and it must manually be deleted to free up space when upgrading node versions, even patch editions. Wouldn't it be simpler just to extract to the `OUT_DIR`? I don't think there will be too many gains since a new node.lib is needed per individual version of nodejs.

If you feel like its necessary, I'll probably change that directory's architecture to be something like `/{target_url_base64}/node-{version}-{arch}.lib` to prevent node and electron files from stomping each other.

Brooooooklyn:
> Are there any real benefits to using a global `node.lib` cache @Brooooooklyn? It seems like a hassle to maintain correctly, it pollutes the home dir, and it must manually be deleted to free up space when upgrading node versions, even patch editions. Wouldn't it be simpler just to extract to the `OUT_DIR`? I don't think there will be too many gains since a new node.lib is needed per individual version of nodejs.
> 
> 
> 
> If you feel like its necessary, I'll probably change that directory's architecture to be something like `/{target_url_base64}/node-{version}-{arch}.lib` to prevent node and electron files from stomping each other.

No I don't think there are any benifit to setup a global node.lib cache. It's inherit from `xray` project . And I want refactor it too.

adumbidiot:
Ok I will have it extract to the `OUT_DIR`.

adumbidiot:
To solve the last 2 issues, the `napi-sys` build script must change somewhat. In order to avoid using the wrong nodejs headers, we simply need to download them from the internet, regardless of the platform, essentially mimicking what is currently done for Windows. 

The bigger issue is determining the correct napi version. I don't believe npm exposes this information through the environment. The only two solutions i can see are including a massive database of nodejs versions and their napi versions, or locating the `#define NAPI_VERSION` in the `node_version.h` file, probably with a regex.

Beyond these two issues, I think the only remaining work is adding a test suite of some kind, though that can potentially be done in another pr as most of the work is done now.

Brooooooklyn (approved):

Issue #277, opened 2020-11-03
Most sys enums are unsafe


adumbidiot:
According to the [bindgen docs for rustified enums](https://docs.rs/bindgen/0.55.1/bindgen/struct.Builder.html#method.rustified_enum):
> Use this with caution, creating this in unsafe code (including FFI) with an invalid value will invoke undefined behaviour. You may want to use the newtype enum style instead.

This is problematic, as NAPI often adds to these enums across editions. This causes UB if, for example, an napi 3 module is loaded in an napi 6 context and calls [typeof](https://nodejs.org/api/n-api.html#n_api_napi_typeof) on a bigint, creating a rust enum with an invalid state. 

This can be fixed by not using rust enums for these types and using constant values instead. Many `From`/`Into` impls will have to be replaced with `TryFrom/TryInto`, but these impls were unsafe anyways. 

I ran into this issue while working on #271. If you want, I can push fixes there or I can just make a new PR fixing this.

Brooooooklyn:
> You may want to use the newtype enum style instead

So how about change bindgen to generate newtype enum?

adumbidiot:
Ok, I'll open a new PR for that

adumbidiot:
@Brooooooklyn This should be reopened. We use low level sys enums directly in napi calls again which can lead to them having an invalid state between versions of napi.

Issue #272, opened 2020-11-01
ThreadSafeFunction is unsafe


adumbidiot:
According to the [docs](https://nodejs.org/api/n-api.html#n_api_reference_counting_of_thread_safe_functions):
> napi_threadsafe_function objects are destroyed when every thread which uses the object has called napi_release_threadsafe_function() or has received a return status of napi_closing in response to a call to napi_call_threadsafe_function

> Upon receiving a return value of napi_closing from napi_call_threadsafe_function() a thread must not use the thread-safe function anymore because it is no longer guaranteed to be allocated.

This library should check if any calls to a threadsafe function result in `napi_closing` in order to prevent a use-after-free bug. 

Maybe we could signal a "poisoned" function by setting the inner `raw_tsfn` to null, and either asserting that it isn't null or returning an error?

Also, what do you think about calling the release function on `drop`? 

This type also look like it used to implement `Clone`. However, this impl was unsafe and looks like it has been removed. Maybe we could reintroduce a manual implementation of clone again, calling [`napi_acquire_threadsafe_function`](https://nodejs.org/api/n-api.html#n_api_napi_acquire_threadsafe_function)? 

In any case, this invalid `Clone` impl makes it fairly easy to cause use-after-frees in the current `crates.io` version of napi, myabe you should consider yanking it?

Brooooooklyn:
> This library should check if any calls to a threadsafe function result in `napi_closing` in order to prevent a use-after-free bug.

If I understood correctly, `napi_closing` status only occurs when `thread_count` to be 0, which only could be decrease by `ThreadsafeFunction::release`, so I don't think we should check whether the return status is `napi_closing`.

> Also, what do you think about calling the release function on drop

Yes, so we don't need provide `acquire/release` any more.

And similarly I think we don't need to provide `refer/unref` too, we just `unref` `ThreadSafeFunction` in `drop`. Let the `ownership/lifetime` system to track these things.

PR #295, opened 2020-11-10
feat(napi): impl Clone and Drop for ThreadSafeFunction, return Status  from ThreadSafeFunction::call

Brooooooklyn:
/cc @adumbidiot 

Close https://github.com/napi-rs/napi-rs/issues/277
Close https://github.com/napi-rs/napi-rs/issues/272

### DIFF
--- a/napi/src/async_work.rs
+++ b/napi/src/async_work.rs
@@ -107,16 +107,16 @@ unsafe extern "C" fn complete<T: Task>(
   match check_status(status).and_then(move |_| value) {
     Ok(v) => {
       let status = sys::napi_resolve_deferred(env, deferred, v.raw());
-      debug_assert!(status == sys::napi_status::napi_ok, "Reject promise failed");
+      debug_assert!(status == sys::Status::napi_ok, "Reject promise failed");
     }
     Err(e) => {
       let status = sys::napi_reject_deferred(env, deferred, e.into_raw(env));
-      debug_assert!(status == sys::napi_status::napi_ok, "Reject promise failed");
+      debug_assert!(status == sys::Status::napi_ok, "Reject promise failed");
     }
   };
   let delete_status = sys::napi_delete_async_work(env, napi_async_work);
   debug_assert!(
-    delete_status == sys::napi_status::napi_ok,
+    delete_status == sys::Status::napi_ok,
     "Delete async work failed"
   );
 }
--- a/napi/src/env.rs
+++ b/napi/src/env.rs
@@ -597,7 +597,7 @@ impl Env {
     R: 'static + Send + FnMut(ThreadSafeCallContext<T>) -> Result<Vec<V>>,
   >(
     &self,
-    func: JsFunction,
+    func: &JsFunction,
     max_queue_size: usize,
     callback: R,
   ) -> Result<ThreadsafeFunction<T>> {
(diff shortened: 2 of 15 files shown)


adumbidiot:
Most of those `debug_assert!`s need to be `asserts`. Otherwise, compiling in release mode will silently use invalid values. Also, calling abort on one function will invalid all of them by sending an `napi_closing` error to the call function, which we need to intercept and track. After receiving this error, the function may be deallocated and should not be used. We should use that `aborted` flag to signal whether `raw_tsfn` is safe to use.

adumbidiot:
I don't think a global atomic bool is necessary here? All we would need to do is make each tsfn clone have an aborted bool, and update it if call returns `napi_closing`. Otherwise, an abort to one function will abort all of them, even unrelated, un-aborted tsfns.

Brooooooklyn:
> I don't think a global atomic bool is necessary here? All we would need to do is make each tsfn clone have an aborted bool, and update it if call returns napi_closing

I wrote a simple test case to test this case:

image

The print status is `Closing`. Which means **close cloned tsfn** will also close all `tsfn` shared the `napi_threadsafe_function`.

adumbidiot:
I was concerned about something like:
```rust
let tsfn1 = ...
let tsfn2 = ...

tsfn1.abort()?;
let status = tsfn.call(Ok(vec![1, 2]), ThreadsafeFunctionCallMode::Blocking);
println!("{:#?}", status); // This should not be aborted since tsfn1 and tsfn2 are two different functions (not clones).
```

Brooooooklyn:
I think `Arc<AtomicBool>` is suitable for this case

adumbidiot:
I'm pretty sure you would still need a guard. Something like:
```rust
let tsfn = ...
let tsfn_clone = tsfn.try_clone().unwrap();

tsfn_clone.abort().unwrap();

// Not Safe
tsfn.call(Ok(vec![1, 2]), ThreadsafeFunctionCallMode::Blocking);
```

Also, calling abort and continuing to use other api functions is unsafe.

```rust
let tsfn = ...
tsfn.abort().unwrap();

// Not Safe
tsfn.call(Ok(vec![1, 2]), ThreadsafeFunctionCallMode::Blocking);
```

Brooooooklyn:
> I'm pretty sure you would still need a guard. For example, the following isn't safe:

Yes, I will add guard in `call` method.

> Also, calling abort and continuing to use other api functions is unsafe:

hmmm, this situation will never happen because `abort` take the ownership and can not call `call` method then.

adumbidiot:
Oops missed that function signature. abort seems fine then.

I think you would need to use guards for all  tsfn functions. For example:
```
let tsfn = ...
let tsfn_clone = tsfn.clone();

tsfn_clone.abort().unwrap();

// Safe due to guard
tsfn.call(Ok(vec![1, 2]), ThreadsafeFunctionCallMode::Blocking);

// Not Safe, lacks guard. 
// Furthermore, raw_tsfn is definitely deallocated here as every function has called abort and/or received napi_closing from call
tsfn.try_clone().unwrap();
```

Brooooooklyn:
`try_clone` already guarded. I think`refer/unref` don't need to be guarded because they can only be called within main thread, and return `Error` if something was wrong.

adumbidiot:
The problem isn't which thread they can be called from, but whether the tsfn still exists.

From the docs:
> Upon receiving a return value of napi_closing from napi_call_threadsafe_function() a thread must not use the thread-safe function anymore because it is no longer guaranteed to be allocated

So we must guard all usages of raw_tsfn

Issue #297, opened 2020-11-13
exit 3221225477 running with nodejs x32 on windows 64 in release mode


Brooooooklyn:
Everything is just fine in debug mode, but node.exe exit with status 1 while loading the `.node` file  compiled with `--release` flag

See https://github.com/napi-rs/package-template/actions/runs/361398266

adumbidiot:
I think this is a compiler codegen error. I don't think working around it is wise as `opt-level='z'` breaks it on my computer again. I found a similar issue: https://github.com/rust-lang/rust/issues/67497. I think we should try to find a smaller example of this behavior and report it to the rust compiler repo.

Brooooooklyn:
Fixed `test_module` in this repo, but failed again in https://github.com/napi-rs/package-template/pull/86/files ....

Brooooooklyn:
Confirmed fixed by change `CString::as_ptr()` to `CString::into_raw()` in windows, which was considered to cause memory leak.

Seems like `node` will free the `CString` and cause double free if using `as_ptr` ?

Brooooooklyn:
wired, still failed in CI https://github.com/napi-rs/package-template/pull/86/checks

But in my windows virtual machine I was using node@10.x 32bit, I will try again tomorrow night.

adumbidiot:
Pointers created from into_raw must be returned to rust and freed with from_raw. If node tries to free these pointers, we have much bigger issues.

adumbidiot:
After debugging, I found that attaching `#[inline(never)]` to `check_status` fixed the issue on my machine. This is definitely a rust codegen issue. That function is inlined incorrectly on that target.

Brooooooklyn:
@adumbidiot oh, let me try

Brooooooklyn:
Seems not work in CI https://github.com/napi-rs/package-template/pull/86/checks?check_run_id=1428438105

adumbidiot:
The latest stable breaks it again

adumbidiot:
Setting `opt-level = 's'` appears to generate good code on my computer, though it will probably break too in the future.

Brooooooklyn:
`opt-level='s'` seems not work on my computer, but opt-level=1 could work

Brooooooklyn:
> opt-level = 1 worked fine

Not work actually, just can execute `require('index.node')`, exit `3221225477 ` again while running tests

pantharshit00:
[...]

PR #298, opened 2020-11-13
fix(napi): windows-i686 release build problem

Brooooooklyn:
Close https://github.com/napi-rs/napi-rs/issues/297

### DIFF
--- a/napi/src/js_values/mod.rs
+++ b/napi/src/js_values/mod.rs
@@ -257,7 +257,7 @@ macro_rules! impl_object_methods {
       where
         T: NapiValue,
       {
-        let key = CString::new(name)?;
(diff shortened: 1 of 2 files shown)


napi/src/js_values/mod.rs
```diff
@@ -257,7 +257,7 @@ macro_rules! impl_object_methods {
       where
         T: NapiValue,
       {
-        let key = CString::new(name)?;
+        let key = CString::new(name.to_owned())?;
```
Brooooooklyn, line 260: Don't know why this line is required.
If without `to_owned`, add `println!("anything")` between `let key....` and `check_status(...` could prevent `3221225477` exit too...

Issue #311, opened 2020-11-19
Some functions leak cstrings


adumbidiot:
Opening this to track issues crated by #309. Some napi functions now leak cstrings.

PR #312, opened 2020-11-20
fixx(napi): CString memory leak

Brooooooklyn:
Close https://github.com/napi-rs/napi-rs/issues/311


adumbidiot:
Got it, thanks

adumbidiot (approved):
Seems good

PR #318, opened 2020-11-23
Refactor napi

Brooooooklyn:
1. inline everything
2. change `check_status` and `type_of` to macro
3. provide #[module_exports] macro
4. remove debug and repr[transparent] for ffi struct

### DIFF
--- a/napi-derive/src/lib.rs
+++ b/napi-derive/src/lib.rs
@@ -1,7 +1,5 @@
 extern crate proc_macro;
 
-use std::mem;
-
 use proc_macro::TokenStream;
 use proc_macro2::{Ident, Literal};
 use quote::{format_ident, quote};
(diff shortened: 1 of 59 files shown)


napi-derive/src/lib.rs
```diff
@@ -30,6 +30,7 @@ struct JsFunction {
   args: Vec<FnArg>,
   name: mem::MaybeUninit<Ident>,
```
adumbidiot: Why isn't `Option` used here for `name`, `signature`, and `signature_raw`?

adumbidiot (approved):
Looks good, I just had one question.

napi-derive/src/lib.rs
Brooooooklyn: Oh, it's because I don't like `unwrap` before, but I think it's better to change to `Option` now

PR #481, opened 2021-02-26
Unwrap from ref

Brooooooklyn:


pimeys:
So this one just removes the need for `try_clone` and now the function pointer implements `Clone`?

Brooooooklyn:
> So this one just removes the need for `try_clone` and now the function pointer implements `Clone`?

Yes

PR #508, opened 2021-03-18
Add a way to convert stateful (Rust) closures into `JsFunction`s.

danielhenrymantilla:
Preview of the mainly added function (under `impl Env`):

Screen Shot 2021-03-18 at 21 04 55

While doing this, I found the `arg_len` parameter in `CallContext::new()` to be redundant (and thus even dangerous, should it end up different to the len of the vec), so I've also taken the liberty of trimming that part (adapting the derive accordingly).

Finally, two nits I've spotted when working with this crate: the `ref_count` parameter of `Ref`'s constructor was not used for the actual construction, and also there is an issue with the `maybe_ref` usage in `object.rs` I haven't fixed that one since I wasn't dealing with that part of the API yet, but I thought it best to at least add a comment warning about it).

### DIFF
--- a/napi-derive/src/lib.rs
+++ b/napi-derive/src/lib.rs
@@ -114,7 +114,7 @@ pub fn js_function(attr: TokenStream, input: TokenStream) -> TokenStream {
       }
 
       let mut env = unsafe { Env::from_raw(raw_env) };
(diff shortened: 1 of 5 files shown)


Brooooooklyn:
Thanks!

> the ref_count parameter of Ref's constructor was not used for the actual construction, and also there is an issue with the maybe_ref usage in object.rs

I will take a look soon

PR #519, opened 2021-03-30
Check compile target at runtime in build

c-nixon:
I ran into some issues cross compiling from macos to x86_64-unknown-linux-gnu. 

The current compile time checks don't work well for this as the build script is always compiled for the host, so the Mac linker args were getting set for a linux build.

This PR moves the check to the build script's runtime, ie the library compile time

### DIFF
--- a/build/Cargo.toml
+++ b/build/Cargo.toml
@@ -10,7 +10,4 @@ repository = "https://github.com/napi-rs/napi-rs"
 version = "1.0.1"
 
 [dependencies]
-cfg-if = "1"
-
-[target.'cfg(windows)'.dependencies]
 ureq = "2"
--- a/build/Cargo.toml
+++ b/build/Cargo.toml
@@ -10,4 +10,7 @@ repository = "https://github.com/napi-rs/napi-rs"
 version = "1.0.1"
 
 [dependencies]
+cfg-if = "1"
+
+[target.'cfg(not(target_env = "musl"))'.dependencies]
 ureq = "2"
--- a/build/src/lib.rs
+++ b/build/src/lib.rs
@@ -1,10 +1,23 @@
 mod macos;
-mod windows;
+
+cfg_if::cfg_if! {
+  if #[cfg(not(target_env = "musl"))] {
+    mod windows;
+  }
+}
 
 pub fn setup() {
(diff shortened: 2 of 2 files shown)


build/Cargo.toml
```diff
@@ -10,7 +10,4 @@ repository = "https://github.com/napi-rs/napi-rs"
 version = "1.0.1"
 
 [dependencies]
-cfg-if = "1"
-
-[target.'cfg(windows)'.dependencies]
```
Brooooooklyn, line 15: I think this line could not be removed for now, because windows target could only be built with `x86_64-pc-windows-msvc` which could not be cross compiled in the other platforms.

c-nixon: It may not be a common practice, but I've compiled rust into a windows exe from Linux as part of CI before, the instructions at https://github.com/est31/msvc-wine-rust work very well

Brooooooklyn: Looks great! But at least we should exclude the `ureq` on Linux-musl and non-x86-x64 host. It`s not possible cross compile windows-msvc on these platform.
And from the CI, install `ureq` failed on Linux-musl host.

c-nixon: Looks like the problem is with the ring dependency of rustls dep in ureq. There is a issue to reinstate native-tls support in ureq over at https://github.com/algesten/ureq/issues/319.

I updated it to exclude the windows code on musl targets. I don't think there'll be many people trying to cross compile to windows from their Raspberry Pi build server using wine and qemu, but hey there might be someone somewhere!

c-nixon: Hmm there seems to be something up with the aarch64-linux action, it's not getting to the build step due to some dependency issues

```
The following packages have unmet dependencies:
 g++-9-aarch64-linux-gnu : Depends: libstdc++-9-dev-arm64-cross (>= 8) but it is not going to be installed
 gcc-9-aarch64-linux-gnu : Depends: libgcc-9-dev-arm64-cross (>= 9.3.0-17ubuntu1~20.04cross2) but it is not going to be installed
                           Recommends: libc6-dev-arm64-cross (>= 2.13-0ubuntu6) but it is not installable
```

PR #523, opened 2021-03-31
Add a method to get argument back after calling js function

### LINKED ISSUE #436: Get mutated argument after function call
I have a `JsObject` created in Rust that I pass in as an argument to `JsFunction.call`. The function mutates the object in JS context. Afterwards, I want to read the mutated object and see what's changed.

Here's the sample code:

```javascript
class XPlugin {
    apply(obj) {
        obj.num = 5;
    }
}

module.exports = {
    plugins: [
        new XPlugin(),
    ]
};
```

```rust
#[js_function(1)]
pub fn bundle(cx: CallContext) -> napi::Result<JsUndefined> {
    let jsconf = cx.get::<JsObject>(0)?; // the module.exports object from above
    if jsconf.has_named_property("plugins")? {
        let plugins = jsconf.get_named_property::<JsArrayBuffer>("plugins")?;
        for i in 0..plugins.get_array_length()? {
            let mut obj = cx.env.create_object()?;
            obj.set_named_property("num", cx.env.create_int32(4)?)?;

            let plugin = plugins.get_element::<JsObject>(i)?;
            let apply_func = plugin.get_named_property::<JsFunction>("apply")?;
            apply_func.call(None, &[obj.into_unknown()])?;

            // Here I want to get the updated obj.num value and print it or something.
        }
    }
    cx.env.get_undefined()
}
```

Is there any way to accomplish this?

kdy1:
Closes #436.

### DIFF
--- a/napi/src/threadsafe_function.rs
+++ b/napi/src/threadsafe_function.rs
@@ -277,17 +277,23 @@ impl<T: 'static> ThreadsafeFunction<T, ErrorStrategy::CalleeHandled> {
   /// See [napi_call_threadsafe_function](https://nodejs.org/api/n-api.html#n_api_napi_call_threadsafe_function)
   /// for more information.
   pub fn call(&self, value: Result<T>, mode: ThreadsafeFunctionCallMode) -> Status {
+    self.call_with_result(value, mode).0
+  }
+
+  pub fn call_with_result(
+    &self,
+    value: Result<T>,
+    mode: ThreadsafeFunctionCallMode,
+  ) -> (Status, Result<T>) {
     if self.aborted.load(Ordering::Acquire) {
-      return Status::Closing;
+      return (Status::Closing, value);
     }
-    unsafe {
-      sys::napi_call_threadsafe_function(
-        self.raw_tsfn,
-        Box::into_raw(Box::new(value)) as *mut _,
-        mode.into(),
-      )
-    }
-    .into()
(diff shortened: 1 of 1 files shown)


Brooooooklyn:
Thanks, but it looks like a breaking change. I will find a backwards compatible way to provide this function.

kdy1:
Is it really breaking change?
I added a method without changing signature of original method, so I'm not sure if it's breaking.

Brooooooklyn:
Oh，seems like tests failed due to other reasons.I'll take a look

napi/src/threadsafe_function.rs
```diff
@@ -277,35 +277,43 @@ impl<T: 'static> ThreadsafeFunction<T, ErrorStrategy::CalleeHandled> {
...
-        self.raw_tsfn,
-        Box::into_raw(Box::new(value)) as *mut _,
-        mode.into(),
-      )
+      return (Status::Closing, value);
     }
-    .into()
+    let data = Box::into_raw(Box::new(value)) as *mut Result<T>;
+    let status =
+      unsafe { sys::napi_call_threadsafe_function(self.raw_tsfn, data as *mut _, mode.into()) }
```
Brooooooklyn, line 293: CI failed because the `value: Result<T>` dropped here, and segfault when take it back `Box::from_raw` after in `call_js_cb`

kdy1:
Closing as it's a wrong approach.

Issue #581, opened 2021-05-27
A memory leak in `JsString::into_utf8`


pimeys:
Hi! A new leak, this time from `JsString`'s `into_utf8` method, see:

```rust
use napi::{CallContext, Env, JsObject, JsString, JsUndefined};
use napi_derive::{js_function, module_exports};

#[js_function(1)]
fn test_sync(ctx: CallContext) -> napi::Result<JsUndefined> {
    let query = ctx.get::<JsString>(0)?.into_utf8()?;
    ctx.env.get_undefined()
}

#[module_exports]
pub fn init(mut exports: JsObject, _env: Env) -> napi::Result<()> {
    exports.create_named_method("test_sync", test_sync)?;

    Ok(())
}
```

And we call this from JS:

```javascript
const api = require(`./libnapi_test.so.node`)

function main() {
    const q = `query {
        findFirstBooking(where: { id: "ckp5jrg9o0000gwj6hlllncqk" }) {
            id
            name
            forename
            description
            email
            phone
            arrivalDate
            departureDate
            price
            advance
            advanceDueDate
            kids
            adults
            status
            nourishment
            createdAt
            room {
                id
                name
            }
        }
    }`;


    let query = {
        query: q,
        variables: {}
    };

    while (true) {
        api.test_sync(JSON.stringify(query));
    }
}
main()
```

This will just skyrocket the memory usage, gigabytes in a few seconds! (at least on this workstation). The leak is gone if you remove `into_utf8` call from the Rust code. Malloc shows `JsString` values getting piled to the heap.

pimeys:
This is fixed by removing the `ManuallyDrop` wrapper in https://github.com/napi-rs/napi-rs/blob/main/napi/src/js_values/string/mod.rs#L65-L67

I'd do a pull request, but I don't know do you need this wrapper to something elsewhere... Memory is flat and stable after patching this out though.

pimeys:
Btw, `from_js_value` too leaks a lot. What we can do now is just stringify input params for the `query` function and deserialize them with serde by ourselves. Having a flat response with my branch and this setup, but not yet sure did my changes break something...

Brooooooklyn:
> Btw, from_js_value too leaks a lot.

Confirmed in my local machine.

> What we can do now is just stringify input params for the query function and deserialize them with serde by ourselves.

[...]

PR #582, opened 2021-05-27
Allow dropping of string&buffer values when converting

pimeys:
The utf8 and utf16 values leak memory a lot, if we don't handle the `ManuallyDrop` values. This fix is first try of changing it. I'd remove the `ManuallyDrop` values completely, but I don't know are they actually needed elsewhere in the code.

Closes: https://github.com/napi-rs/napi-rs/issues/581

### DIFF
--- a/napi/src/js_values/string/utf16.rs
+++ b/napi/src/js_values/string/utf16.rs
@@ -1,6 +1,6 @@
-use std::convert::TryFrom;
 use std::mem;
 use std::ops::Deref;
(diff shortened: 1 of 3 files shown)


Brooooooklyn:
Looks good to me! I can't remember why `JsStringUtf8` and `JsStringUtf16` implemented with `ManuallyDrop`, but I'm sure that they shoule be drop while dropping the `JsStringUtf8` and `JsStringUtf16`.
Thanks!

PR #629, opened 2021-06-26
fix: 🧪 Rust enum to JS value serialization

### LINKED ISSUE #507: Serializing Rust enum to JS value loses variant information
This:

```
#[derive(Serialize, Deserialize)]
enum MyEnum {
    A { value: u32 },
    B { value: u32 },
}

pub fn call(ctx: CallContext) -> Result<JsUnknown> {
    ctx.env.to_js_value(&MyEnum::A { value: 15 })
}
```

Results in:

```json
{ "value": 15 }
```

But should result in ([see Serde](https://serde.rs/enum-representations.html))

```json
{ "A": { "value": 15 } }
```

Otherwise there will be no way to know what Enum variant was returned.

Beraliv:
Fixes #507

- [x] Test
- [ ] Implement logic

### DIFF
--- a/test_module/__test__/serde/ser.spec.ts
+++ b/test_module/__test__/serde/ser.spec.ts
@@ -32,3 +32,9 @@ test('serialize make_bytes_struct', (t) => {
     map: 'source map',
   })
 })
+
+test('makes enum from rust to js', (t) => {
+  t.deepEqual(bindings.make_enum(), {
+    A: { value: 15 },
+  })
+})
--- a/test_module/__test__/serde/ser.spec.ts
+++ b/test_module/__test__/serde/ser.spec.ts
@@ -38,7 +38,9 @@ test('makes empty enum from rust to js', (t) => {
 })
 
 test('makes tuple enum from rust to js', (t) => {
-  t.deepEqual(bindings.make_tuple_enum(), [1, '2'])
+  t.deepEqual(bindings.make_tuple_enum(), {
+    Tuple: [1, '2'],
+  })
 })
 
 test('makes struct enum from rust to js', (t) => {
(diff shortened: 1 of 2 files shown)


Beraliv:
@Brooooooklyn 

As I understand, for struct enum we don't create an object, but rather serialise the key as a string:

```rs
fn serialize_unit_variant(
    self,
    _name: &'static str,
    _variant_index: u32,
    variant: &'static str,
  ) -> Result<Self::Ok> {
    self.0.create_string(variant).map(|string| string.0)
}
```

If I try to create an object for it, but I do something different instead:

1. The change:

```rs
#[inline]
fn serialize_unit_variant(
  self,
  _name: &'static str,
  _variant_index: u32,
  variant: &'static str,
) -> Result<Self::Ok> {
  let mut outer = self.0.create_object()?;
  let inner = self.0.create_object()?;
  outer.set_named_property(
    variant,
    JsObject(Value {
      env: inner.0.env,
      value: inner.0.value,
      value_type: ValueType::Object,
    }),
  )?;
  Ok(outer.0)
}
```

2. The result empty enum test case:

```rs
make_test!(make_empty_enum, TypeEnum::Empty);
```

```diff
- {
-   Empty: {},
- }
+ 'Empty'
```

Am I on the right way? Maybe I need to know something, still didn't get everything

test_module/__test__/serde/ser.spec.ts
```diff
@@ -32,3 +32,23 @@ test('serialize make_bytes_struct', (t) => {
     map: 'source map',
   })
 })
+
+test('makes empty enum from rust to js', (t) => {
+  t.deepEqual(bindings.make_empty_enum(), 'Empty')
+})
+
+test('makes tuple enum from rust to js', (t) => {
+  t.deepEqual(bindings.make_tuple_enum(), [1, '2'])
```
Beraliv: todo: change to object here

Beraliv: Done ✅

Beraliv:
I think I will close it as I cannot continue working on it

PR #823, opened 2021-11-01
feat(napi): implement `Either` type

Brooooooklyn:
carbon(6)


crates/napi/src/bindgen_runtime/js_values/either.rs
```diff
@@ -0,0 +1,318 @@
+use super::{FromNapiValue, ToNapiValue, TypeName};
+use crate::{type_of, Status, ValueType};
+
+#[derive(Debug, Clone, Copy)]
+pub enum Either<
```
forehalo, line 7: macro?

Brooooooklyn: I've tried macro, it could not reduce codes from the `from_napi_value` parts, so I gave up.

Brooooooklyn:
What if  `Either<Buffer, ...>` since the result of `type_of!(buffer)` is `Object`

PR #849, opened 2021-11-13
feat(napi): await Promise<T> in async fn

Brooooooklyn:


crates/napi/src/bindgen_runtime/js_values/promise.rs
```diff
@@ -0,0 +1,102 @@
...
+      napi_sys::napi_get_named_property(env, napi_val, CString::new("then")?.as_ptr(), &mut then,),
+      "Failed to get then function"
+    )?;
+
+    let mut then_js_cb = ptr::null_mut();
+    let (tx, rx) = channel();
+    check_status!(
+      napi_sys::napi_create_function(
+        env,
+        CString::new("then")?.as_ptr(),
```
messense: This will cause UB I think?  Looks like the example in [std docs](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.as_ptr)

```rust
use std::ffi::CString;

let ptr = CString::new("Hello").expect("CString::new failed").as_ptr();
unsafe {
    // `ptr` is dangling
    *ptr;
}
```

Brooooooklyn: Yes right, I'll fix it

PR #851, opened 2021-11-14
Fully qualify all types in generated code

### LINKED ISSUE #850: `#[napi(object)]` results in compilation error if `Object` type is imported
I'm guessing the macro expands into something that clashes with imported types?

The following code:
```rust
use goblin::Object;
use napi::bindgen_prelude::*;
use napi_derive::napi;
use serde::{Deserialize, Serialize};
use std::{fs, path::Path};

#[napi(object)]
#[derive(Serialize, Deserialize, Debug)]
pub struct BinaryInfo {
    pub platform: String,
    pub arch: String,
}

#[napi]
pub fn get_info(path: String) -> Result<BinaryInfo> {
    let path = Path::new(&path);
    let buffer = fs::read(path)?;

    match Object::parse(&buffer).map_err(|e| Error::new(Status::GenericFailure, e.to_string()))? {
        _ => {}
    }

    Ok(BinaryInfo {
        platform: "".to_string(),
        arch: "".to_string(),
    })
}

```

Results in:
```
error[E0599]: the variant or associated item `from_napi_value` exists for enum `goblin::Object<'_>`, but its trait bounds were not satisfied
   --> src\lib.rs:7:1
    |
7   | #[napi(object)]
    | ^^^^^^^^^^^^^^^ variant or associated item cannot be called on `goblin::Object<'_>` due to unsatisfied trait bounds
    |
   ::: C:\Users\tim\.cargo\registry\src\github.com-1ecc6299db9ec823\goblin-0.4.3\src\lib.rs:295:5
    |
295 |     pub enum Object<'a> {
    |     -------------------
    |     |
    |     doesn't satisfy `_: napi::bindgen_prelude::FromNapiValue`
    |     doesn't satisfy `goblin::Object<'_>: NapiValue`
    |
    = note: the following trait bounds were not satisfied:
            `goblin::Object<'_>: NapiValue`
            which is required by `goblin::Object<'_>: napi::bindgen_prelude::FromNapiValue`
            `&goblin::Object<'_>: NapiValue`
            which is required by `&goblin::Object<'_>: napi::bindgen_prelude::FromNapiValue`
            `&mut goblin::Object<'_>: NapiValue`
            which is required by `&mut goblin::Object<'_>: napi::bindgen_prelude::FromNapiValue`
    = note: this error originates in the attribute macro `napi` (in Nightly builds, run with -Z macro-backtrace for more info)
```

If I rename `Object`, (ie. `use goblin::Object as Obj;`), compilation succeeds.

timfish:
Fixes #850 

Could some of the types in `napi::bindgen_prelude` now be moved elsewhere to keep the prelude smaller?

For example, would `ctor` or `___CALL_FROM_FACTORY` ever be used by `napi` users, or were they only in the prelude for use in generated code?

### DIFF
--- a/crates/backend/src/codegen/enum.rs
+++ b/crates/backend/src/codegen/enum.rs
@@ -34,7 +34,7 @@ impl NapiEnum {
     });
 
     quote! {
-      impl TypeName for #name {
+      impl napi::bindgen_prelude::TypeName for #name {
         fn type_name() -> &'static str {
           #name_str
         }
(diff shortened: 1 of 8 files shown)


Brooooooklyn:
Great!

Brooooooklyn (approved):

PR #871, opened 2021-11-22
Fix cargo:rerun-if commands

sam3d:
The `rerun-if-changed` command expects a path, but as these are env vars (and the path `./TYPE_DEF_TMP_PATH` doesn't exist), this re-runs the build script every time. This PR fixes this and uses `rerun-if-env-changed` instead

https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorerun-if-env-changedname

### DIFF
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -2,8 +2,8 @@ mod macos;
 mod windows;
 
 pub fn setup() {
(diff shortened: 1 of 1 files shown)


Brooooooklyn:
This is really helpful, thank you!

sam3d:
You're welcome, thank you for creating such an awesome library!

PR #919, opened 2021-12-10
feat(napi): support external deref

h-a-n-a:
Support External Value DeRef 
```
*external.as_ref() -> *external
```


Brooooooklyn:
We need `DerefMut` too

h-a-n-a:
> We need `DerefMut` too

done :)

PR #938, opened 2021-12-18
feat(napi): refine return type

liulinboyi:
Refine return type


forehalo:
@liulinboyi thanks for your contribution :)

PR #947, opened 2021-12-19
Add `skip_typescript` attribute for macro

h-a-n-a:
This PR basically adds `skip_typescript` functionality to `napi` macro, which helps to skip some type declarations that we might not want to expose to users, especially when we use `External` to create some cache for later use.

## Examples:

**enums:**
```rust
#[napi(skip_typescript)]
enum NinjaTurtle {
  Leo,
  Raph
}
```

**struct:**
```rust
#[napi]
struct NinjaTurtle {
   pub name: String
  
   #[napi(skip_typescript)]
   pub mask_color: String
}

impl NinjaTurtle {
   #[napi(skip_typescript)]
   pub fn new_leo() -> Self {
       Self { name: "Leo".to_owned(), mask_color: "Blue".to_owned() }
   }
}
```

**fn:**
```rust
#[napi(skip_typescript)]
fn xx() {}
```

**const:**
```rust
#[napi(skip_typescript)]
const SOME_VALUE: u32 = 1
```


cli/src/build.ts
```diff
@@ -361,7 +361,10 @@ async function processIntermediateTypeFile(
     return idents
   }
 
-  const dtsHeader = `/* eslint-disable */
+  const dtsHeader = `/* tslint:disable */
+/* eslint-disable */
+
+/* auto-generated by Napi-rs */
```
Brooooooklyn: **NAPI-RS**

Issue #407, opened 2021-01-05
napi watch


Brooooooklyn:
Provide `napi watch` command to watch and rebuild napi package.

Brooooooklyn:
Closed in https://github.com/napi-rs/napi-rs/pull/1492

PR #952, opened 2021-12-20
chore: cargo build --watch

stone-jin:
#407

usage: 
```
$ cargo build --watch
```

cargo build --help echo:
```
 --watch       watch and rebuild napi package
```

### DIFF
--- a/cli/src/build.ts
+++ b/cli/src/build.ts
@@ -37,7 +37,7 @@ export class BuildCommand extends Command {
   })
 
   isWatch = Option.Boolean(`--watch`, false, {
-    description: 'watch and rebuild napi package'
+    description: 'watch and rebuild napi package',
   })
 
   isRelease = Option.Boolean(`--release`, false, {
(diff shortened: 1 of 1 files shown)


Brooooooklyn:
@stone-jin Thank you for contributing. I'll merge it after 2.0.1 is released. Because it's a big feature and I want to do some more works top on it.

stone-jin:
got it. @Brooooooklyn

forehalo:
Maybe we should rewrite the cli tool in rust as well, so it can work with [cargo-watch](https://crates.io/crates/cargo-watch) out of box.
And also the current cli tool written in node smells bad because of the progressing of migrating to esm modules. I can't believe we need to bundle node program. 😞
IMO, it's good time to redesign it.

Brooooooklyn:
Yes, we should rewrite the cli in `Rust` also because of
- https://github.com/napi-rs/napi-rs/issues/981

stone-jin:
🙋‍♂️ support it

forehalo:
well, I think it would be fine to close this PR because of #1074, you may want to contribute to that PR if you might hurry to use the `watch` ability

Issue #982, opened 2021-12-29
Add a contributing guide


messense:
For someone who  isn't following the ever changing Nodejs ecosystem, it's hard to setup a local development environment for contributing to napi-rs.🤪

PR #983, opened 2021-12-29
Add support for cross compilation with Zig

messense:
https://actually.fyi/posts/zig-makes-rust-cross-compilation-just-work/

So that you don't need to compile in a docker container.

Example:

```bash
❯ uname -a
Darwin m1.lan 21.2.0 Darwin Kernel Version 21.2.0: Sun Nov 28 20:29:10 PST 2021; root:xnu-8019.61.5~1/RELEASE_ARM64_T8101 arm64

messense in m1 in napi-rs/examples/napi on  zig is 📦 v0.1.0 via  v17.3.0 via 🦀 v1.57.0
❯ yarn run build --zig --target x86_64-unknown-linux-gnu
yarn run v1.22.17
$ node ../../cli/scripts/index.js build --js false --zig --target x86_64-unknown-linux-gnu
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
✨  Done in 0.67s.

messense in m1 in napi-rs/examples/napi on  zig is 📦 v0.1.0 via  v17.3.0 via 🦀 v1.57.0
❯ file index.node
index.node: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, with debug_info, not stripped
```

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -47,6 +47,7 @@
     "chalk": "4",
     "clipanion": "^3.1.0",
     "debug": "^4.3.3",
+    "env-paths": "^3.0.0",
     "fdir": "^5.1.0",
     "inquirer": "^8.2.0",
     "js-yaml": "^4.1.0",
--- a/cli/src/build.ts
+++ b/cli/src/build.ts
@@ -1,16 +1,17 @@
 import { execSync } from 'child_process'
-import { existsSync } from 'fs'
+import { existsSync, mkdirSync, writeFileSync } from 'fs'
 import { join, parse, sep } from 'path'
 
 import { Instance } from 'chalk'
 import { Command, Option } from 'clipanion'
+import envPaths from 'env-paths'
 import { groupBy } from 'lodash-es'
 import toml from 'toml'
 
 import { getNapiConfig } from './consts'
 import { debugFactory } from './debug'
 import { createJsBinding } from './js-binding-template'
-import { getDefaultTargetTriple, parseTriple } from './parse-triple'
+import { getCpuArch, getDefaultTargetTriple, parseTriple } from './parse-triple'
 import {
   copyFileAsync,
   mkdirAsync,
@@ -117,6 +118,10 @@ export class BuildCommand extends Command {
     required: false,
   })
 
+  useZig = Option.Boolean(`--zig`, false, {
+    description: `Use ${chalk.green('zig')} as linker`,
+  })
+
   async execute() {
     const cwd = this.cargoCwd
       ? join(process.cwd(), this.cargoCwd)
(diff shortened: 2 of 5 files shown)


cli/src/build.ts
```diff
@@ -160,6 +165,25 @@ export class BuildCommand extends Command {
...
     }
+
+    if (this.useZig && triple.platform == 'linux') {
+      const paths = envPaths('napi-rs')
+      const cpuArch = getCpuArch(triple.arch)
+      const linkerWrapper = join(paths.cache, `zig-cc-${triple.abi}.sh`)
+      mkdirSync(paths.cache, { recursive: true })
+      writeFileSync(
+        linkerWrapper,
+        `#!/bin/bash\nzig cc \${@/-lgcc_s/-lunwind} -target ${cpuArch}-linux-${triple.abi}\n`,
```
messense: Obviously this don't work on Windows yet.

cli/src/parse-triple.ts
```diff
@@ -21,6 +21,14 @@ const CpuToNodeArch: { [index: string]: NodeJSArch } = {
   armv7: 'arm',
 }
 
+const NodeArchToCpu: { [index: string]: string } = {
+  arm64: 'aarch64',
+  ppc: 'powerpc',
+  ppc64: 'powerpc64',
```
messense, line 27: I'm not sure what endianness are `ppc` and `ppc64` on Nodejs.

cli/package.json
```diff
@@ -47,6 +47,7 @@
     "chalk": "4",
     "clipanion": "^3.1.0",
     "debug": "^4.3.3",
+    "env-paths": "^3.0.0",
```
messense, line 50: This package is kinda an arbitrary choice, not sure if there's a better one.

cli/src/build.ts
```diff
@@ -160,6 +165,25 @@ export class BuildCommand extends Command {
...
+        { mode: 0o700 },
+      )
+      const envTarget = triple.raw.replaceAll('-', '_').toUpperCase()
+      Object.assign(additionalEnv, {
+        TARGET_CC: linkerWrapper,
+        TARGET_CXX: 'zig c++',
```
messense: Need to pass `-target` and maybe other args to `zig c++` as well.

yisibl:
Interesting

messense:
What do you think? @Brooooooklyn

Brooooooklyn:
@messense it's great. But should we set up some CI to test it? Or just mark it as experimental?

messense:
> But should we set up some CI to test it?

Definitely, I just don't know how to do it properly for this project yet. #982 could be helpful.

> Or just mark it as experimental?

For now we should mark it as experimental since it's not supported on Windows yet.

Brooooooklyn:
@messense I'll add a contribute guide tomorrow! Thanks for your help.

messense:
Added a linux aarch64 musl cross compiling with ziglang test workflow.

Brooooooklyn:
Let's merge it and do some test top on it.

cli/src/build.ts
messense: Windows support is a little trickier, see https://github.com/PyO3/maturin/blob/36e525bba838eb10af324ec8805bf715a2b37139/src/zig.rs#L51-L57

PR #1017, opened 2022-01-08
bindgen_runtime: impl FromNapiValue for serde_json::Number

### LINKED ISSUE #1013: Unable to call a rust method from JS with a `number` if the method argument has type `serde_json::Value`
[The docs](https://napi.rs/docs/concepts/function#arguments) imply that the rust type `serde_json::Value` corresponds to the JS/TS type `unknown`, which includes `number`. However I'm getting this runtime error when I try to call such a method with a number:
https://github.com/napi-rs/napi-rs/blob/349f747ce47218f4fcece19130e6898746a4651e/crates/napi/src/bindgen_runtime/js_values/serde.rs#L41-L44

I assume this is a deliberate decision because you don't know whether the passed number was a float or int or uint or bigint. However, it would be nice to have **any way** to specify the default behavior instead of simply erroring. Maybe a feature flag or something so we can say _"I don't care what the number in JS is, make it a float"_.

zeroslope:
fix #1013


AlCalzone:
What's the easiest way to test this locally?

AlCalzone:
Hmm, I'm still getting the error `Js Number is not be able to convert to rust`. Seems that `napi` is picking up a different version than my lib.

Brooooooklyn:
Oh my bad, you should also change the `napi` version to the same GitHub url & branch

AlCalzone:
This is working for me. 👍🏻

I have a couple of questions/thoughts though:
1. Should integers be detected and converted automatically (or potentially opt-in with a flag)? For example, `JSON.stringify(1.0)` results in `'1'` while doing the same through `napi-rs/serde` now results in `'1.0'`.
2. Alternatively, should there be a flag to specify the default Rust type for numbers - as opposed to always `f64`?
3. While we're at the topic of differences between JS/Rust (probably something for a followup PR):
	```
	> JSON.stringify(() => {})
	undefined
	> JSON.stringify(undefined)
	undefined
	```
	yet the same via `napi-rs/serde` results in `'null'` both times.

AlCalzone:
OT, but might be interesting:
As expected, passing a value to Rust, letting it JSON-serialize and passing it back to JS is slower than `JSON.stringify` in JS. I wanted to know just how much slower... On my machine roughly 5..7 million op/s vs. 10 million.

Except 32 bit numbers with the high bit set, here Rust is ~5% faster (5.5 vs 5.2 million op/s):
canvas

Brooooooklyn:
@AlCalzone really interesting, I got similar results in https://github.com/napi-rs/napi-rs/blob/main/bench/get-array-from-js.ts#L11

Brooooooklyn:
@AlCalzone BTW, have you tried https://docs.rs/napi/latest/napi/struct.Env.html#method.from_js_value ?

Brooooooklyn:
> yet the same via napi-rs/serde results in 'null' both times.

I think `napi-rs/serde` should follow up the `JSON.stringify` behavior in this scenario, since there is no `null` primitive in `JSON`.

AlCalzone:
> @AlCalzone BTW, have you tried https://docs.rs/napi/latest/napi/struct.Env.html#method.from_js_value ?

I haven't yet but I can try

> I think `napi-rs/serde` should follow up the `JSON.stringify` behavior in this scenario, since there is no `null` primitive in `JSON`.

There is, but only `null`  itself serializes to `'null'`:
```
Welcome to Node.js v16.10.0.
Type ".help" for more information.
> JSON.stringify(null)
'null'
> JSON.parse('null')
null
```

AFAIK the problem is that the default in `from_napi_value` is `Null` and not `()`.

AlCalzone:
FYI, I've opened https://github.com/zeroslope/napi-rs/pull/1 onto @zeroslope's fork which fixes the things I mentioned before. Only `null` is now represented as `Null`, `undefined`, functions, symbols and everything else that's unknown will throw errors.

AlCalzone:
@Brooooooklyn What are your plans with this? Is this something you're planning to merge?
I've started to build on this PR, e.g. with automatic number conversion (https://github.com/AlCalzone/napi-rs/commits/auto-number), so further upstream updates are going to get messy 😅

Brooooooklyn:
@AlCalzone I'd like to merge this if you finished number conversion

AlCalzone:
Ok, I'll raise a separate PR since I can't push to the current one.

AlCalzone:
PR is at https://github.com/napi-rs/napi-rs/pull/1052, but I'll have to rebase.

PR #1020, opened 2022-01-11
fix(napi): abort threadsafe functions upon env cleanup

dceddia:
This fixes a deadlock/panic in Electron when the window is reloaded. See #1021 and the comments below.

### DIFF
--- a/crates/napi/src/threadsafe_function.rs
+++ b/crates/napi/src/threadsafe_function.rs
@@ -211,9 +211,13 @@ impl<T: 'static, ES: ErrorStrategy::T> ThreadsafeFunction<T, ES> {
       )
     })?;
 
+    let aborted = Arc::new(AtomicBool::new(false));
+    let aborted_ptr = Arc::into_raw(aborted.clone()) as *mut c_void;
+    check_status!(unsafe { sys::napi_add_env_cleanup_hook(env, Some(cleanup_cb), aborted_ptr) })?;
+
     Ok(ThreadsafeFunction {
       raw_tsfn,
-      aborted: Arc::new(AtomicBool::new(false)),
+      aborted,
       ref_count: Arc::new(AtomicUsize::new(initial_thread_count)),
       _phantom: PhantomData,
     })
@@ -321,6 +325,11 @@ impl<T: 'static, ES: ErrorStrategy::T> Drop for ThreadsafeFunction<T, ES> {
   }
 }
 
+unsafe extern "C" fn cleanup_cb(cleanup_data: *mut c_void) {
+  let aborted = unsafe { Arc::<AtomicBool>::from_raw(cleanup_data.cast()) };
+  aborted.store(true, Ordering::SeqCst);
+}
+
 unsafe extern "C" fn thread_finalize_cb<T: 'static, V: NapiRaw, R>(
   _raw_env: sys::napi_env,
   finalize_data: *mut c_void,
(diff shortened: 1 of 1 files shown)


crates/napi/src/threadsafe_function.rs
```diff
@@ -211,9 +211,13 @@ impl<T: 'static, ES: ErrorStrategy::T> ThreadsafeFunction<T, ES> {
       )
     })?;
 
+    let aborted = Arc::new(AtomicBool::new(false));
+    let aborted_ptr = Arc::into_raw(aborted.clone()) as *mut c_void;
+    check_status!(unsafe { sys::napi_add_env_cleanup_hook(env, Some(cleanup_cb), aborted_ptr) })?;
```
dceddia, line 216: An env cleanup hook is registered, and in the cleanup callback, aborted is set true so that ThreadsafeFunction will stop making use of the underlying tsfn handle

crates/napi/src/threadsafe_function.rs
```diff
@@ -341,6 +350,11 @@ unsafe extern "C" fn call_js_cb<T: 'static, V: NapiRaw, R, ES>(
   R: 'static + Send + FnMut(ThreadSafeCallContext<T>) -> Result<Vec<V>>,
   ES: ErrorStrategy::T,
 {
+  // env and/or callback can be null when shutting down
+  if raw_env.is_null() || js_callback.is_null() {
+    return;
+  }
```
dceddia, line 356: When V8 cleans up, it clears the call queue by calling the registered callback (`call_js_cb`) with null values for the `env` and `js_callback` args. Added a check to return early in this case.

crates/napi/src/threadsafe_function.rs
```diff
@@ -404,10 +418,9 @@ unsafe extern "C" fn call_js_cb<T: 'static, V: NapiRaw, R, ES>(
       sys::napi_get_and_clear_last_exception(raw_env, &mut error_result),
       sys::Status::napi_ok
     );
-    assert_eq!(
-      sys::napi_fatal_exception(raw_env, error_result),
-      sys::Status::napi_ok
-    );
+    // When shutting down, napi_fatal_exception sometimes returns another exception
+    let stat = sys::napi_fatal_exception(raw_env, error_result);
+    assert!(stat == sys::Status::napi_ok || stat == sys::Status::napi_pending_exception);
```
dceddia, line 430: Occasionally, V8 seems to start its env cleanup in the middle of a call to `call_js_cb`, and the JS function call will return with a status of `napi_pending_exception`. When that happens, it seems to get "stuck" in the napi_pending_exception state, and attempting to raise an exception with `napi_fatal_exception()` will return `napi_pending_exception` instead of `napi_ok`. Adjusted the assertion to handle this case.

Brooooooklyn, line 430: It's really weird, I mean it seems like a bug of Node.js itself.

dceddia: It does seem weird! Looking at the Electron code for `napi_fatal_exception`, it really looks like the only thing it can return is `napi_ok`, so either the code I'm looking at isn't what's running, or somewhere it's setting the exception again.

dceddia: I guess trying to raise an exception during shutdown doesn't make much sense, though? I could see why it might be like "hey it's too late for that, your exception failed". But it seems weird that it would be so intermittent. I really had to mash the refresh button a lot to get this one to happen 😂

Brooooooklyn:
@dceddia could you please rebase the latest main branch? I think your implementation is good for now.

dceddia:
@Brooooooklyn All set

dceddia:
Whoops, forgot clippy, just fixed it. Boy, those github build failure emails really come in a flood haha.

Brooooooklyn:
Don't worry about the `clippy` errors, I've fixed them in https://github.com/napi-rs/napi-rs/pull/1029

Issue #1022, opened 2022-01-11
Strings are not converted when Object::get::<_,&str> is called


Tpt:
`Object::get::<_,String>` works perfectly and properly converts the string from UTF-16 to UTF-8 but `Object::get::<_,&str>` does not seem to run the conversion and returns an invalid `str`.

For example the following function crashes when called:

```rust
use napi::Env;
use napi_derive::napi;

#[napi]
fn test(env: Env) {
    let mut obj = env.create_object().unwrap();
    obj.set("test", "foo").unwrap();
    assert_eq!(Some("foo"), obj.get("test").unwrap());
}
```

With the error:
```
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `Some("foo")`,
 right: `Some("_�")`', nodejs/src/lib.rs:10:5
fatal runtime error: failed to initiate panic, error 5
```

messense:
Looks like a use-after-free bug to me.

https://github.com/napi-rs/napi-rs/blob/80c97b2660be34412425629451a9a794f1264774/crates/napi/src/bindgen_runtime/js_values/string.rs#L85-L100

The `ret` Vec is dropped when out of scope but a `CStr` is constructed from a pointer to it and return to caller.

Brooooooklyn:
I'm wondering how to avoid such issues by `miri`. It seems hard to integrate `miri` with libraries like `napi-rs` which has a lot of `ffi` calls.

messense:
Maybe we can try asan first? https://github.com/japaric/rust-san#addresssanitizer

messense:
BTW, maybe we should also consider adding `#![forbid(unsafe_op_in_unsafe_fn)]` lint to make unsafe blocks smaller: https://rust-lang.github.io/rfcs/2585-unsafe-block-in-unsafe-fn.html

PR #1023, opened 2022-01-12
fix(napi): invalid memory address in FromNapiValue for &str

Brooooooklyn:
Fix: 
- https://github.com/napi-rs/napi-rs/issues/1022


crates/napi/src/bindgen_runtime/js_values/string.rs
```diff
@@ -84,14 +84,32 @@ impl FromNapiValue for &str {
...
       sys::napi_get_value_string_utf8(env, napi_val, buf_ptr, len, &mut written_char_count),
       "Failed to convert napi `string` into rust type `String`"
     )?;
 
+    // The `&str` should only be accepted from function arguments.
+    // We shouldn't implement `FromNapiValue` for it before.
+    // When it's used with `Object.get` scenario, the memory which `&str` point to will be invalid.
+    // For this scenario, we create a temporary empty `Object` here and assign the `Vec<u8>` under `&str` to it.
+    // So we can safely forget the `Vec<u8>` here which could fix the memory issue here.
+    // This implementation should be removed in next major release.
```
messense:
```suggestion
    // FIXME: This implementation should be removed in next major release.
```

PR #1025, opened 2022-01-13
fix(cli): missing zig-abi-suffix support

Brooooooklyn:


messense:
Nice!

I think abi suffix only works with glibc now, last time I tried with musl it doesn't work(musl has 1.1 and 1.2 abi versions), so it's better to ignore or report error when it's specified with musl targets.

Brooooooklyn:
> so it's better to ignore or report error when it's specified with musl targets.

I'll note it should only work with `gnu` in the next release.

PR #1026, opened 2022-01-13
feat(cli): add `--strip` option for removing symbols

messense:
For [this example](https://github.com/napi-rs/napi-rs/tree/main/examples/napi)

Without `--strip`
```bash
❯ ls -lrth index.node
-rwxr-xr-x  1 messense  staff   5.1M Jan 13 16:59 index.node
```

With `--strip`
```bash
❯ ls -lrth index.node
-rwxr-xr-x  1 messense  staff   1.9M Jan 13 16:59 index.node
```

### DIFF
--- a/cli/src/build.ts
+++ b/cli/src/build.ts
@@ -162,6 +162,10 @@ export class BuildCommand extends Command {
     )} ${chalk.green('--zig-abi-suffix=2.17')}`,
   })
 
+  isStrip = Option.Boolean(`--strip`, false, {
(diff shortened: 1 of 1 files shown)


Brooooooklyn:
Does it compatible with all linkers or just zig/lld  ?

messense:
> 

It's compatible with all linkers. See https://github.com/rust-lang/cargo/issues/3483#issuecomment-431209957

Brooooooklyn:
Great!

PR #1027, opened 2022-01-13
fix(cli): properly handle RUSTFLAGS env var

messense:
Fixes https://github.com/yisibl/resvg-js/pull/47


messense:
I think we should consider using `cargo rustc` instead of `cargo build` so we don't need to mess with `RUSTFLAGS`.

https://stackoverflow.com/a/60578951

PR #1036, opened 2022-01-17
chore: run unit tests with asan

messense:


messense:
Here's an example: https://github.com/messense/napi-rs/runs/4837807932?check_suite_focus=true

```rust
#[napi]
fn contains(source: String, target: String) -> bool {
  let xs = [0, 1, 2, 3];
  let _y = unsafe { *xs.as_ptr().offset(4) };
  source.contains(&target)
}
```

.github/workflows/test.yaml
```diff
@@ -79,6 +79,25 @@ jobs:
...
+          profile: minimal
+          override: true
+
+      - name: Unit tests with address sanitizer
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          yarn build:test
+          LD_PRELOAD=/usr/lib/gcc/x86_64-linux-gnu/9/libasan.so yarn test
+        env:
+          RUST_TARGET: x86_64-unknown-linux-gnu
```
messense: Maybe cli should use the `CARGO_BUILD_TARGET` env var instead.

https://doc.rust-lang.org/cargo/reference/environment-variables.html#configuration-environment-variables

.github/workflows/test.yaml
```diff
@@ -79,6 +79,25 @@ jobs:
...
+          RUST_BACKTRACE: 1
+          RUSTFLAGS: -Z sanitizer=address
+          ASAN_OPTIONS: detect_leaks=0
```
messense: Too many false positives with leak sanitizer turned on.

Brooooooklyn:
Wow, great!

messense:
Re-run to see if the issue persists.

Brooooooklyn (approved):

PR #1043, opened 2022-01-20
fix(napi-derive): an `Option` in front of a required parameter is no …

JoostK:
…longer optional in TypeScript

This fixes a bug where having a parameter of type `Option<T>` followed by required parameters
would incorrectly declare the parameter as option in the TypeScript declaration file, resulting
in invalid syntax.


JoostK:
I was looking for tests but couldn't find any. If someone can point me to the tests then I'll add some.

JoostK:
Ah, this project is used for tests: https://github.com/napi-rs/napi-rs/tree/main/examples/napi/src.

I'll add some there.

Brooooooklyn:
Great!

PR #1048, opened 2022-01-23
Feat/get js function

Brooooooklyn:


crates/napi/src/bindgen_runtime/module_register.rs
```diff
@@ -138,6 +160,57 @@ pub fn register_class(
...
+      Some(JsFunction(Value {
+        env: *env,
+        value: function,
+        value_type: ValueType::Function,
+      }))
+    })
+    .ok_or_else(|| {
+      crate::Error::new(
+        crate::Status::InvalidArg,
+        "JavaScript function not existed".to_owned(),
```
mohsen1: “JavaScript function <name> does not exists” reads better

Brooooooklyn: Thanks

PR #1052, opened 2022-01-25
feat: impl `FromNapiValue` for `serde_json::Number`, fix it for `Null`, throw for impossible types

AlCalzone:
This PR supersedes https://github.com/napi-rs/napi-rs/pull/1017
I've extended it with the following extra changes:

- Only `null` is now represented as `Null`.
- `undefined`, `functions`, `symbols `and everything else that's unknown will throw errors when attempting to convert to a `serde_json::Value`.
- For JS numbers, the appropriate Rust type (f64, u32 or i32) is automatically chosen. This ensures that JSON serialization in Rust results in the same output as in JS.


AlCalzone:
@Brooooooklyn I'm not sure how to run and update these test snapshots. `yarn build:test ; yarn test` succeeds.

Brooooooklyn:
`yarn test -u`

AlCalzone:
Ok looks like my base branch wasn't 100% up to date

PR #1056, opened 2022-02-04
Fix String roundtrip with interior nul bytes

### LINKED ISSUE #1055: JS strings get null-terminated during a roundtrip
I can't yet put my finger on where this happens, but JS strings may contain NULL characters, which seem to terminate the string when passing to napi-rs. Consider the following scenario:
```rs
// lib.rs

#[napi]
pub fn echo(echo: String) -> String {
  echo
}
```

```js
// javascript:
const assert = require("assert");
const {echo} = require("./lib");

const expected = "what up?!\u0000after the NULL";
const actual = echo(expected);
assert.strictEqual(actual, expected); // This throws, b/c actual is "what up?!"
process.exit(0);
```

messense:
Fixes #1055


AlCalzone:
Tested manually on my use case - this works!

Issue #1059, opened 2022-02-04
Cannot run on Node 14


bontibon:
I am trying to use the `--zig` flag with @napi-rs/cli@2.4.2 and Node 14. The code path, however, includes a String `.replaceAll` call which is not implemented, causing this error:

```
Type Error: linkerWrapper.replaceAll is not a function
    at BuilderCommand.<anonymous> (...scripts/index.js:12808)
```

PR #1067, opened 2022-02-08
fix(cli): compatible for Node.js 10x

Brooooooklyn:
Close https://github.com/napi-rs/napi-rs/issues/1059


messense:
I think we should also run zig tests with Node 12 on CI to make sure it won't be broken in the future.

PR #1070, opened 2022-02-10
chore(napi): switch from windows crate to windows-sys

messense:
> The windows-sys crate is a zero-overhead fallback for the most demanding situations and primarily where the absolute best compile time is essential. It only includes function declarations (externs), structs, and constants. No convenience helpers, traits, or wrappers are provided.

https://github.com/microsoft/windows-rs#windows-sys


messense:
Looks like napi requires some of the helpers, traits, or wrappers. Closing.

Issue #1061, opened 2022-02-04
CLI: Allow specifying the location where the *.node file is output


Xaeroxe:
The end user may want the *.node file to go somewhere other than the `CARGO_MANIFEST_DIR`. Sure they can copy the file out after the fact, but it'd be nice if they didn't have to.

Brooooooklyn:
Closed in https://github.com/napi-rs/napi-rs/pull/1492

Issue #1045, opened 2022-01-22
Feature Request: CLI option to include/exclude debug symbols on all platforms


timfish:
I might want to include debug symbols for full backtrace support on all platforms.

Currently there is `--strip` but this only removes symbols on unix'y platforms. There no way to include symbols in output artifacts on Windows.

Would it be better to instead have an option where the value did the same thing across all platforms? 

- Enabled, all platforms would include symbols
- Disabled, all platforms would include symbols

This would potentially be a breaking change because the current default is inconsistent across platforms.

messense:
AFAIK, debug symbols aren't not consistent across platforms either, on Linux they can be embed in ELF files, on macOS they're saved in separated `.dSYM` folders, I'm not sure about Windows.

Thus there are some complications, for example, how to distribute debug symbols in NPM packages.

Issue #988, opened 2022-01-02
New project created with @napi-rs/cli doesn't create .cargo dir, .js, .d.ts files


AlCalzone:
I tried to follow the getting started guide at https://napi.rs/docs/introduction/simple-package but I'm getting stuck pretty early. According to the documentation, the generated directory should contain a `.cargo` directory and building should generate `index.js` and `index.d.ts` files in the root. Neither happens for me.

Even after changing the build commands to
```
    "build": "napi build --platform --release --js index.js --dts index.d.ts",
    "build:debug": "napi build --platform --js index.js --dts index.d.ts",
```
`yarn build` does not generate these files.

What might I be missing?

Brooooooklyn:
@AlCalzone did you install the latest `@napi-rs/cli` in global?

AlCalzone:
Yeah, I did. Just minutes before I tried that.

JMLX42:
@AlCalzone what version of `yarn` are you using?

Running `yarn run build` with yarn 1.22.17 properly creates the `index.d.ts` and `index.js` files. But with yarn 3.2.0 the files are *not* generated and there is an error in the console:

```
Failed to write type def file: Os { code: 20, kind: NotADirectory, message: "Not a directory" }
```

Brooooooklyn:
@JMLX42 Do you have the pnp feature enabled?

JMLX42:
> Do you have the pnp feature enabled?

@Brooooooklyn how can I check that? What is the recommended config?

JMLX42:
I have the following in my `.yarnrc.yml`:

```
nodeLinker: node-modules
```

Which AFAIK means PNP is disabled. Is that what napi expects?

AlCalzone:
> what version of `yarn` are you using?

I didn't change anything from the template. It also seemed to work on another project, so I'm at a loss what was wrong there.

Brooooooklyn:
Close in https://github.com/napi-rs/napi-rs/pull/1492

Issue #981, opened 2021-12-29
A more generic cli implementation


ztytotoro:
Since napi-rs will support both `node` and `deno`(and also `WebAssembly`), but it's a bit weird that using npm cli to build a deno project.

Would you consider rewriting cli by rust or something else. Maybe we can

1. use rust to rewrite cli, and wrap it with `napi-rs` to publish to both `npm` and `deno` registry
2. use deno, it's very easy to write a cli tool with deno, and use [`deno/dnt`](https://github.com/denoland/dnt) to build npm package

PR #1074, opened 2022-02-14
brand new cli tools written in rust

forehalo:
the working/tracing PR of refactoring napi cli tools with rust.

features: 
- [x] napi new  
- [x] napi build
- [x]  napi watch
- [ ] 🚘napi xxx

Aimed to fix all known issues with current cli as well:
- [x] #981
- [x] #988 
- [x] #407
- [x] #1061
- [x] #1170 
- [x] #1188 
- [ ] #1189 
- [ ] #1045


crates/cli/src/commands/build/mod.rs
```diff
@@ -0,0 +1,523 @@
...
+      "CARGO_TARGET_{}_LINKER",
+      self.target.triple.to_uppercase().replace('-', "_")
+    );
+
+    let envs = [("CC", &cc), ("CXX", &cxx), (linker_env_name.as_str(), &cc)];
+    trace!("set envs for zig cross compiling");
+    for (name, value) in envs.iter() {
+      trace!("  {}: {}", &name, &value.display());
+      cmd.env(name, value);
+    }
```
messense: The env vars are better aligned with `cargo-zigbuild`: https://github.com/messense/cargo-zigbuild/blob/9f31dcb97d5cff618d60b3c1d18dfb9f51045b7c/src/build.rs#L184-L233. 

And there are some hacks we do in `cargo-zigbuild` for some targets: https://github.com/messense/cargo-zigbuild/blob/9f31dcb97d5cff618d60b3c1d18dfb9f51045b7c/src/build.rs#L239-L287

forehalo: The complexity of using `cargo-zigbuild` is much more than my expectation.

I would migrate the old logic from js version which makes it work first and well design the crossing compiling part later on.

messense: Well, it's much easier if you do it this way: https://github.com/PyO3/maturin/blob/922459f19d5a1ba1295401f6b9c13198bcf61c2b/src/compile.rs#L202-L221 , all the env vars and operating system dependency hacks are taken care of.

crates/cli/index.js
```diff
@@ -0,0 +1,3 @@
+const { run } = require('./index.node')
```
Brooooooklyn: Should we publish it as standalone binary?

forehalo: it exists only for test convenience

forehalo:
Unfortunately, the rust cli implementation has to be deprecated because of the complexity of using.

We would soon introduce a redesigned cli in node with both command line and programatic interface.

Issue #1073, opened 2022-02-13
Error when using factory functions inside of multiple Worker threads


dguenther:
## Description

I've noticed intermittent errors when using factory functions inside Node.js Workers. They show up as the following error:

```
node:internal/event_target:777
  process.nextTick(() => { throw err; });
                           ^
Error: Class contains no `constructor`, can not new it!
    at Object.<anonymous> (D:\repos\napi-rs-issue\test.js:14:26)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at MessagePort.<anonymous> (node:internal/main/worker_thread:187:24)
    at MessagePort.[nodejs.internal.kHybridDispatch] (node:internal/event_target:562:20)
    at MessagePort.exports.emitMessage (node:internal/per_context/messageport:23:28)
Emitted 'error' event on Worker instance at:
    at Worker.[kOnErrorMessage] (node:internal/worker:289:10)
    at Worker.[kOnMessage] (node:internal/worker:300:37)
    at MessagePort.<anonymous> (node:internal/worker:201:57)
    at MessagePort.[nodejs.internal.kHybridDispatch] (node:internal/event_target:562:20)
    at MessagePort.exports.emitMessage (node:internal/per_context/messageport:23:28)
```

These errors don't seem to happen when using `#[napi(constructor)]`. I'm fairly sure it's a concurrency-related issue, because it seems to only happen in Workers. I created an example reproduction case in the repository below to make it easy to test changes.

I'm not too familiar napi code, but I'm wondering if it's something to do with this: https://github.com/napi-rs/napi-rs/blob/14773af1598cfc4cfe5fa6fd61768c78cc542f99/crates/napi/src/bindgen_runtime/callback_info.rs#L93

## Reproduction

I created a reproduction case on the `factory-threads` branch here: https://github.com/dguenther/napi-rs-issues/tree/factory-threads

PR #1081, opened 2022-03-01
fix(napi): race issues with Node.js worker_thread

Brooooooklyn:
Close https://github.com/napi-rs/napi-rs/issues/1073

### DIFF
--- a/crates/napi/src/bindgen_runtime/js_values/number.rs
+++ b/crates/napi/src/bindgen_runtime/js_values/number.rs
@@ -1,4 +1,5 @@
 use super::{check_status, sys, Result};
+use crate::type_of;
 
 macro_rules! impl_number_conversions {
 	( $( ($name:literal, $t:ty, $get:ident, $create:ident) ,)* ) => {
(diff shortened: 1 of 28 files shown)


crates/napi/src/bindgen_runtime/js_values/number.rs
```diff
@@ -43,8 +44,9 @@ macro_rules! impl_number_conversions {
...
           check_status!(
             unsafe { sys::$get(env, napi_val, &mut ret) },
-						"Failed to convert napi value into rust type `{}`",
-            $name
+						"Failed to convert napi value {:?} into rust type `{}`",
+            type_of!(env, napi_val),
+            $name,
           )?;
 
 					Ok(ret)
```
janpio: Replaces tabs with spaces:
```suggestion
            "Failed to convert napi value {:?} into rust type `{}`",
            type_of!(env, napi_val),
            $name,
          )?;

            Ok(ret)
```
(Extension "Refined Github" makes these visible in PRs)

Issue #1126, opened 2022-04-11
Fails to build with Rust v1.57.0


rkeene:
I'm attempting to use this with Rust v1.57.0, however it currently fails:

```
$ npm run build

> @ build /home/rkeene/devel/node/local/lib/asn1rs
> napi build --release --target x86_64-unknown-linux-gnu

   Compiling napi v2.2.0
error[E0658]: use of unstable library feature 'cstring_from_vec_with_nul'
   --> /home/rkeene/.cargo/registry/src/github.com-1ecc6299db9ec823/napi-2.2.0/src/bindgen_runtime/module_register.rs:387:31
    |
387 |           let js_class_name = CString::from_vec_with_nul_unchecked(js_name.as_bytes().to_vec());
    |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #73179 <https://github.com/rust-lang/rust/issues/73179> for more information

For more information about this error, try `rustc --explain E0658`.
error: could not compile `napi` due to previous error
Internal Error: Command failed: cargo build --release --target x86_64-unknown-linux-gnu
    at checkExecSyncError (child_process.js:790:11)
    at Object.execSync (child_process.js:863:15)
    at BuildCommand.<anonymous> (/home/rkeene/devel/node/local/lib/asn1rs/node_modules/@napi-rs/cli/scripts/index.js:14602:26)
    at Generator.next (<anonymous>)
    at fulfilled (/home/rkeene/devel/node/local/lib/asn1rs/node_modules/@napi-rs/cli/scripts/index.js:3423:58)

```

```sh
$ dpkg -s rustc
Package: rustc
Status: install ok installed
...
Version: 1.57.0+dfsg1+llvm-0ubuntu1~20.04.1
```

Brooooooklyn:
Closed in https://github.com/napi-rs/napi-rs/pull/1130

PR #1127, opened 2022-04-12
change MSRV to `1.58.0`

devtomio:
closes https://github.com/napi-rs/napi-rs/issues/1126

[`CString::from_vec_with_nul_unchecked`](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.from_vec_with_nul_unchecked) was only stabilized in `1.58.0`


Brooooooklyn:
Close in https://github.com/napi-rs/napi-rs/pull/1130

Issue #1124, opened 2022-04-09
Fatal error: Check failed: result.second.


mischnic:
I get a native crash when returning an empty buffer for the second time:
```
<Buffer >


#
# Fatal error in , line 0
# Check failed: result.second.
#
#
#
#FailureMessage Object: 0x7ff7b74b9370
 1: 0x108b5adca node::NodePlatform::GetStackTracePrinter()::$_3::__invoke() [/usr/local/Cellar/node@16/16.14.2/bin/node]
 2: 0x10935c4c1 V8_Fatal(char const*, ...) [/usr/local/Cellar/node@16/16.14.2/bin/node]
 3: 0x108e48244 v8::internal::GlobalBackingStoreRegistry::Register(std::__1::shared_ptr<v8::internal::BackingStore>) [/usr/local/Cellar/node@16/16.14.2/bin/node]
 4: 0x108c4976b v8::ArrayBuffer::GetBackingStore() [/usr/local/Cellar/node@16/16.14.2/bin/node]
 5: 0x108b2415e node::ArrayBufferViewContents<char, 64ul>::Read(v8::Local<v8::ArrayBufferView>) [/usr/local/Cellar/node@16/16.14.2/bin/node]
 6: 0x108ae441f void node::Buffer::(anonymous namespace)::StringSlice<(node::encoding)5>(v8::FunctionCallbackInfo<v8::Value> const&) [/usr/local/Cellar/node@16/16.14.2/bin/node]
 7: 0x108c851e3 v8::internal::FunctionCallbackArguments::Call(v8::internal::CallHandlerInfo) [/usr/local/Cellar/node@16/16.14.2/bin/node]
 8: 0x108c84c14 v8::internal::MaybeHandle<v8::internal::Object> v8::internal::(anonymous namespace)::HandleApiCallHelper<false>(v8::internal::Isolate*, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::FunctionTemplateInfo>, v8::internal::Handle<v8::internal::Object>, v8::internal::BuiltinArguments) [/usr/local/Cellar/node@16/16.14.2/bin/node]
 9: 0x108c843f3 v8::internal::Builtin_HandleApiCall(int, unsigned long*, v8::internal::Isolate*) [/usr/local/Cellar/node@16/16.14.2/bin/node]
[...]

PR #1144, opened 2022-04-23
fix(napi): use null pointer for empty buffers

mischnic:
Use actual `0x0` pointers for empty buffers, this is also what Node itself does (when creating an empty buffer in JS, passing it to native code and printing the data pointer).

Related to https://github.com/napi-rs/napi-rs/issues/1124

This has come up once again with Parcel: https://github.com/DoctorEvidence/lmdb-js/issues/161

The added check could also be put into a utility function that would be called like this in the various methods: `null_ptr_if_empty(len, data as *mut c_void)` 

---

https://github.com/napi-rs/napi-rs/issues/1124#issuecomment-1094585366:
> After thinking about it for a while, I think this problem is hard to avoid at the NAPI-RS. Empty Vec case is easy to prevent, but there are many other cases that may provide the same pointer for different buffers.

Even if this might doesn't fix every case, returning an empty vec is easy to do even unintentionally and also hard to work around when returning a struct containing a buffer via serde.


Brooooooklyn (approved):
Thank you!

Issue #1146, opened 2022-04-25
Returned empty buffer is invalid: `Cannot perform %TypedArray%.prototype.set on a detached ArrayBuffer`


mischnic:
The empty buffers returned are somehow still (after https://github.com/napi-rs/napi-rs/pull/1144) invalid (using Node 16):
```
/Users/niklas/Desktop/napi-rs-empty-buffer/x.js:16
buffer.set(broken);
       ^

TypeError: Cannot perform %TypedArray%.prototype.set on a detached ArrayBuffer
    at Buffer.set (<anonymous>)
    at Object.<anonymous> (index.js:16:8)
```

On Node 14 and 16, both a JS-created empty Buffer and the buffer from napi-rs  have the same data pointer (0), length and `.constructor`. Verified using 
```cpp
Napi::Value print(const Napi::CallbackInfo &info) {
	Napi::Env env = info.Env();
	Napi::Buffer<uint8_t> arg0 = info[0].As<Napi::Buffer<uint8_t>>();
	Napi::Buffer<uint8_t> arg1 = info[1].As<Napi::Buffer<uint8_t>>();
	std::cout << "a " << (long)arg0.Data() << ", length: " << arg0.ElementLength()
			  << ", size: " << (int)arg0.ElementSize() << "\n";
	std::cout << "b " << (long)arg1.Data() << ", length: " << arg1.ElementLength()
			  << ", size: " << (int)arg1.ElementSize() << "\n";
	return env.Undefined();
}
```

On Node 12, no error is thrown but the data pointer of the JS-created buffer is some (random) pointer and not 0.

### Reproduction

Using
```rust
#[napi]
fn transform() -> Buffer {
  Buffer::from(vec![])
}
```
```toml
napi = {version = "2.3.2", default-features = false, features = ["napi4"]}
```
and then
```js
const mod = require("./test.darwin-x64.node");

let working = Buffer.from([]);
let broken = mod.transform();

let shared = new ArrayBuffer(0);
let buffer = Buffer.from(shared);

buffer.set(broken);
buffer.set(working);

console.log(shared);
```

mischnic:
The same happens if I do `Napi::Buffer<uint8_t>::New(env, nullptr, 0)`, but `Napi::Buffer<uint8_t>::New(env, 0)` works.

Internally, only the first one uses `napi_create_external_buffer` but the second one uses `napi_create_buffer`:

[...]

PR #1147, opened 2022-04-26
fix(napi): use create_buffer/arrary_buffer if provided data is empty

Brooooooklyn:
Close https://github.com/napi-rs/napi-rs/issues/1146


mischnic:
Thank you for the quick fix!

PR #1163, opened 2022-05-03
feat(wasm): experimental support for wasm32-unknown-unknown

Brooooooklyn:
- [ ] Support JavaScript Primitives types
- [ ] Support `wasm32-unknown-emscripten` target
- [ ] Update Test/Publish workflows

### DIFF
--- a/.github/workflows/wasm.yml
+++ b/.github/workflows/wasm.yml
@@ -1,0 +1,57 @@
+name: wasm32-unknown-unknown
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+env:
+  DEBUG: 'napi:*'
+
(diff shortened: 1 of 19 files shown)


jcrugzz:
@Brooooooklyn just wanted to note wasm32-wasi may be the more appropriate target to use given thats where the future spec is going. (we use wasm32-wasi to build envoy wasm plugins rather than wasm32-unknown-unknown

d3lm:
wasm32-wasi would definitely be more appropriate long term, but wasm32-unknown-unknown would be more universal, for now at least, for environments that don't yet fully support WASI.

d3lm:
In case it help, WebContainers now support WASI.

PR #1165, opened 2022-05-06
feat(napi): support obtaining node return value for threadsafe function 

h-a-n-a:


examples/napi/src/threadsafe_function.rs
```diff
@@ -3,61 +3,158 @@ use std::thread;
...
 pub fn call_threadsafe_function(callback: JsFunction) -> Result<()> {
   let tsfn: ThreadsafeFunction<u32, ErrorStrategy::CalleeHandled> = callback
-    .create_threadsafe_function(0, |ctx| {
-      ctx.env.create_uint32(ctx.value + 1).map(|v| vec![v])
-    })?;
-  for n in 0..100 {
+    .create_threadsafe_function(
+      0,
+      |ctx| ctx.env.create_uint32(ctx.value + 1).map(|v| vec![v]),
+      |res: Promise<u32>| {},
```
forehalo: `res: Promise<T>` is dropped immediately after promise get constructed, so the channel inside it is closed as well. 

```
// RUST, simplified
promise.then((v) => { 
  // `.then` callback will be getting called in next loop
  // so the cpu keeps going on the rust side, till here(commented line), where the created promise dropped
  sender.send(v)
})`
```

h-a-n-a: Thank you for the quick notice!

h-a-n-a:
For status quo, calling threadsafe function is pure a enqueue procedure. Is it necessary for NAPI-RS to create a unique ID internallly? If the answer is yes, we may need to expose a method to obtain the unique ID generated  by us.

cc @Brooooooklyn @forehalo

h-a-n-a:
As what I discussed with @Brooooooklyn privately, the function for obtaining values from node side should be designed as a separate one from the main `ThreadsafeFunction::call`, or it will become a breaking change to NAPI-RS. So I decide to implement a new `register_js_result_callback` function for each tsfn instance.

cc @forehalo

h-a-n-a:
close for stale

PR #1195, opened 2022-05-24
feat(napi) Implement `ToNapiValue` for 8 and 16-bits numbers

Hywan:
`napi` only supports numbers in 16 and 32-bits. To support Rust
numbers with 8 and 16-bits, we can cast them to 32-bits. To convert
from 8 and 16-bits to 32-bits, we use their `Into` implementation. To
convert from 32-bits to 8 and 16-bits, we use their `TryInto`, we
should theorically not fail, but in case of, an `except` message is
appended.

The code has also been re-indented as it was containing a mix of 2
spaces, 4 spaces and tabs identation.

### DIFF
--- a/crates/napi/src/bindgen_runtime/js_values/number.rs
+++ b/crates/napi/src/bindgen_runtime/js_values/number.rs
@@ -2,8 +2,8 @@ use super::{check_status, sys, Result};
 use crate::type_of;
 
 macro_rules! impl_number_conversions {
-	( $( ($name:literal, $t:ty, $get:ident, $create:ident) ,)* ) => {
-		$(
+  ( $( ($name:literal, $t:ty as $st:ty, $get:ident, $create:ident) ,)* ) => {
+    $(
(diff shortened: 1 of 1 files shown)


Hywan:
I've not found the tests for that part of the code. However, I've tested in my project and it seems to work.

crates/napi/src/bindgen_runtime/js_values/number.rs
```diff
@@ -49,16 +50,20 @@ macro_rules! impl_number_conversions {
             $name,
           )?;
 
-            Ok(ret)
-				}
+          Ok(ret.try_into().expect(concat!("Failed to convert ", stringify!($st), " to ", stringify!($t))))
```
messense: Since this method already returns a `Result`, should this return an `Err` instead of panicking when conversion failed?

Hywan: Done.

Issue #1222, opened 2022-07-05
Returning a reference to an `Uint8Array`


Hywan:
Hello,

I've the following code:

```rust
#[napi]
pub struct EncryptedAttachment {
    media_encryption_info: …,
    encrypted_data: Uint8Array,
}

#[napi]
impl EncryptedAttachment {
    … 

    #[napi(getter)]
    pub fn encrypted_data(&self) -> Uint8Array {
        Uint8Array::new(self.encrypted_data.deref().to_owned())
    }
}
```

Here, `EncryptedAttachment.encrypted_data` returns a **copy** of the `Uint8Array`. However, in some case, this copy can be costly. Is it possible to avoid the copy, maybe by returning a reference (I tried `Reference`, `Ref` and `&Uint8Array` with no promising result yet)?

Thanks for your help.

PR #1224, opened 2022-07-06
feat(napi): clone reference for TypedArray/Buffer

Brooooooklyn:
- Close https://github.com/napi-rs/napi-rs/issues/1222

```rust
#[napi]
pub struct Context {
  pub buffer: Uint8Array,
  private_fields: ...
}

#[napi]
impl Context {
  #[napi(constructor)]
  pub fn new(buffer: Uint8Array) -> Self {
    Self { buffer, ... }
  }
}
```

The `buffer` could be accessed by `context.buffer` in JavaScript by default:

```ts
const buf = new Uint8Array([1, 2, 3, 4])
const context = new Context(buf)

console.assert(buf === context.buffer)
```


JonathanWilbur:
Does this mean that I have to return a value as a field within an object to avoid cloning it into Javascript-land?

Issue #1220, opened 2022-07-05
Memory and safety issues with `ThreadsafeFunction`


SevInf:
While debugging memory leak in Prisma, we stumbled upon couple of issues with `ThreadsafeFunctions`, which, I beleive, boil down to a not quite correct underlying C APIs usages.

## `unref` leaks memory

`napi-rs` implements its own reference counter on top of `napi_ref_threadsafe_function` and `napi_unref_threadsafe_function ` and skips `napi_release_threadsafe_function` call if it is 0. This was done to fix #518. However, there are couple of issues with that fix:
- despite what their name might suggest, original C APIs are not reference counted. Docs say as much:
   >Similar to uv_ref it is also idempotent.
- they don't really have anything to do with resource management, they just allow event loop to run not waiting for the function cleanup.
  > Neither does napi_unref_threadsafe_function mark the thread-safe functions as able to be destroyed nor does napi_ref_threadsafe_function prevent it from being destroyed.

So, skipping `release` call if unref is called causes the function and all its context to leak. You can see it in my [reproduction repo](https://github.com/SevInf/napi-rs-repro) if you run `npm run build && npm run leaking`: memory usage will steadily climb until the process use up all available heap space and crash.

Neon solved the issue, similar to #518 in [another way](https://github.com/neon-bindings/neon/pull/744). They use `threadsafe_function` finalization callback to mark the function as finalized and skip `release` call in `drop` in that case. `finalize` callback is called either when the last thread releases the function, `abort` is called or environment is destroyed. Solving the issue this way allows to use `unref` without leaking any memory, but if we stop there, we'll have another problem.

## using threadsafe function after finalization callback have been called is not safe

This have not been a problem before, since function never was finalized until VM is destroyed. Now, however, it is possible that call will happen while vm is being destroyed, which is a problem, since the function is not guaranteed to be allocated anymore. Same is true  in situation where `napi_call_threadsafe_function` returns `napi_closing`. In order to avoid segfault in this case, we have to make sure that no `napi_*_threadsafe_function` APIs are called after any of those two events happened.

[...]

PR #1230, opened 2022-07-07
fix(napi): memory leak in ThreadsafeFunction

Brooooooklyn:

@SevInf I can confirm this can fix the memory leak issue in your repro repo, could you try it in the Prisma engine?

### DIFF
--- a/crates/napi/src/threadsafe_function.rs
+++ b/crates/napi/src/threadsafe_function.rs
@@ -5,7 +5,7 @@ use std::ffi::CString;
 use std::marker::PhantomData;
 use std::os::raw::c_void;
 use std::ptr;
-use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
 use crate::bindgen_runtime::ToNapiValue;
@@ -147,7 +147,6 @@ type_level_enum! {
 pub struct ThreadsafeFunction<T: 'static, ES: ErrorStrategy::T = ErrorStrategy::CalleeHandled> {
   raw_tsfn: sys::napi_threadsafe_function,
   aborted: Arc<AtomicBool>,
-  ref_count: Arc<AtomicUsize>,
   _phantom: PhantomData<(T, ES)>,
 }
 
@@ -164,7 +163,6 @@ impl<T: 'static, ES: ErrorStrategy::T> Clone for ThreadsafeFunction<T, ES> {
     Self {
       raw_tsfn: self.raw_tsfn,
       aborted: Arc::clone(&self.aborted),
-      ref_count: Arc::clone(&self.ref_count),
       _phantom: PhantomData,
     }
   }
@@ -196,6 +194,13 @@ impl<T: 'static, ES: ErrorStrategy::T> ThreadsafeFunction<T, ES> {
     let initial_thread_count = 1usize;
     let mut raw_tsfn = ptr::null_mut();
     let ptr = Box::into_raw(Box::new(callback)) as *mut c_void;
+    let aborted = Arc::new(AtomicBool::new(false));
+    let aborted_ptr = Arc::into_raw(aborted.clone()) as *mut c_void;
+    // `aborted_ptr` is passed into both `finalize_callback` and `env_cleanup_callback`.
+    // So increase strong count here to prevent it from being dropped twice.
+    unsafe {
+      Arc::increment_strong_count(aborted_ptr);
+    }
     check_status!(unsafe {
       sys::napi_create_threadsafe_function(
         env,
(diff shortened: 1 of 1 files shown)


crates/napi/src/threadsafe_function.rs
```diff
@@ -196,6 +194,13 @@ impl<T: 'static, ES: ErrorStrategy::T> ThreadsafeFunction<T, ES> {
     let initial_thread_count = 1usize;
     let mut raw_tsfn = ptr::null_mut();
     let ptr = Box::into_raw(Box::new(callback)) as *mut c_void;
+    let aborted = Arc::new(AtomicBool::new(false));
+    let aborted_ptr = Arc::into_raw(aborted.clone()) as *mut c_void;
+    // `aborted_ptr` is passed into both `finalize_callback` and `env_cleanup_callback`.
```
SevInf, line 199: I think you might not need env cleanup callback anymore - `finalize_callback` is called on env cleanup. Right now, one `Arc<AtomicBool>` for every threadsafe function will be kept in memory until env is destroyed, which is sub-optimal. Without env_cleanup_callback, you can let it drop when the `ThreadsafeFunction` drops.

crates/napi/src/threadsafe_function.rs
```diff
@@ -328,18 +336,20 @@ impl<T: 'static, ES: ErrorStrategy::T> Drop for ThreadsafeFunction<T, ES> {
...
 unsafe extern "C" fn thread_finalize_cb<T: 'static, V: ToNapiValue, R>(
   _raw_env: sys::napi_env,
   finalize_data: *mut c_void,
-  _finalize_hint: *mut c_void,
+  finalize_hint: *mut c_void,
 ) where
   R: 'static + Send + FnMut(ThreadSafeCallContext<T>) -> Result<Vec<V>>,
 {
   // cleanup
   drop(unsafe { Box::<R>::from_raw(finalize_data.cast()) });
```
SevInf, line 350: `finalize_data` is no longer points to `Box<callback>`, so I think threadsafe function context is now leaking.

SevInf:
Thanks, @Brooooooklyn, this improves things a lot for us. However, i think there is still some room for improvement. I think this fix, unfortunately, introduces some smaller leak on native side of things, see my comments.

crates/napi/src/threadsafe_function.rs
Brooooooklyn: The env cleanup callback is for electon window reload https://github.com/napi-rs/napi-rs/pull/1020

crates/napi/src/threadsafe_function.rs
```diff
@@ -298,20 +302,24 @@ impl<T: 'static> ThreadsafeFunction<T, ErrorStrategy::Fatal> {
...
+    let status = unsafe {
       sys::napi_call_threadsafe_function(
         self.raw_tsfn,
         Box::into_raw(Box::new(value)) as *mut _,
         mode.into(),
       )
     }
-    .into()
+    .into();
+    if status == Status::Closing {
```
SevInf, line 313: Unfortinately, I think, `AtomicBool` won't cut it here, you'll need a proper lock. `ThreadsafeFunction::call` can be called concurrently, and it is possible that the second call will happen after you called `napi_call_threadsafe_function`, but before you checked the result. In case you received `napi_closing` as the result of the first call, doing the second call is already unsafe.
My original reproduction repo used `RwLock` here, but we later switched to `Mutex` for exactly this reason. If you need it, you can find updated version [here](https://github.com/prisma/prisma-engines/blob/9e513c70f3a00c2f3ebf3115f2545ff2de8bce20/query-engine/query-engine-node-api/src/log_callback.rs).

crates/napi/src/threadsafe_function.rs
Brooooooklyn: No, `finalize_data` is still pointed to `Box<Callback>`, this PR just change `finalize_hint` from `Box<Callback>` to `aborted_ptr`

SevInf: Ah, sorry, my bad

crates/napi/src/threadsafe_function.rs
SevInf: I think `finalize_callback` will take care of that as well.

crates/napi/src/threadsafe_function.rs
Brooooooklyn: I use `Atomic` here rather than `Mutex` is because there is already `Mutex` in the `napi_call_threadsafe_function` in the Node.js; I don't want to introduce redundant overhead here.

Let me extrapolate the memory ordering here. I'll be back here when I have a conclusion 👀

SevInf: Mutex inside `napi_call_threadsafe_function` won't protect you from the issue I am describing though: it will be released right when `napi_call_threadsafe_function` exits, where to avoid unsafe call you have to hold a lock until you verified the status.

PR #1232, opened 2022-07-08
feat: add experimental `Env::get_module_file_name` method

### LINKED ISSUE #1229: Expose `__dirname` to users of napi
The `__dirname` value is important when trying to retrieve bundled assets. As far as I can tell there's no way to access this with `napi` currently, unless you write a JS layer on top of the rust module

docs https://nodejs.org/api/modules.html#__dirname

ErichDonGubler:
Hi there! Coworker of the reporter of https://github.com/napi-rs/napi-rs/issues/1229 (hi @Xaeroxe!) here. Title says all -- I'm bubbling up the experimental `node_api_get_module_file_name` function to the `Env` API. :)

Resolves #1229's use case; we'll probably have to do additional work on our side to actually get the same data.

### DIFF
--- a/crates/napi/src/env.rs
+++ b/crates/napi/src/env.rs
@@ -638,6 +638,25 @@ impl Env {
     Ok(unsafe { JsFunction::from_raw_unchecked(self.0, raw_result) })
   }
 
+  /// This API retrieves the file path of the currently running JS module as a URL. For a file on
+  /// the local file system it will start with `file://`.
+  ///
+  /// TODO: Determine the encoding of the `CStr` provided.
+  ///
+  /// # Errors
+  ///
+  /// The retrieved string may be empty if the add-on loading process fails to establish the
+  /// add-on's file name.
+  #[cfg(feature = "experimental")]
+  pub fn get_module_file_name(&self, f: impl FnOnce(Result<&std::ffi::CStr>)) {
+    let mut char_ptr = ptr::null();
+    let status = unsafe { sys::node_api_get_module_file_name(self.0, &mut char_ptr) };
+    // SAFETY: This is safe because `char_ptr` is guaranteed to not be `null`, and point to
+    // null-terminated string data.
+    let res = check_status!(status).map(|()| unsafe { std::ffi::CStr::from_ptr(char_ptr) });
+    f(res)
+  }
+
   /// This API retrieves a napi_extended_error_info structure with information about the last error that occurred.
   ///
   /// The content of the napi_extended_error_info returned is only valid up until an n-api function is called on the same env.


crates/napi/src/env.rs
```diff
@@ -638,6 +638,25 @@ impl Env {
     Ok(unsafe { JsFunction::from_raw_unchecked(self.0, raw_result) })
   }
 
+  /// This API retrieves the file path of the currently running JS module as a URL. For a file on
+  /// the local file system it will start with `file://`.
+  ///
+  /// TODO: Determine the encoding of the `CStr` provided.
```
ErichDonGubler, line 644: Just one `TODO` here -- I believe this is UTF-8, but I'm not positive.

crates/napi/src/env.rs
```diff
@@ -638,6 +638,25 @@ impl Env {
...
+  ///
+  /// # Errors
+  ///
+  /// The retrieved string may be empty if the add-on loading process fails to establish the
+  /// add-on's file name.
+  #[cfg(feature = "experimental")]
+  pub fn get_module_file_name(&self, f: impl FnOnce(Result<&std::ffi::CStr>)) {
```
ErichDonGubler, line 651: Design note: because the lifetime of the string is tied to the environment, and we don't model the lifetime of the environment even via `Env`, we basically have to (ab)use closures this way in order to expose a safe API.

Brooooooklyn: maybe `'static` is safe enough for this scenario?

Brooooooklyn: I'd like to make it return `Option<'static str>` directly, and return `None` if the underlying `*const c_char` is not a valid  utf8 string.
Then expose another API to return the raw `CStr`.
So that the `get_module_file_name` is enough for most of cases, and the `get_module_file_name_raw` can be used in `filepath` contains invalidate `utf8` string.

Xaeroxe: Or we could just make the whole problem go away by giving a `CString` back. Not sure this code will ever be on a hot path.

ErichDonGubler: @Brooooooklyn:

> maybe `'static` is safe enough for this scenario?

Maybe? 🤔 That would be safe _if_ the lifetime of the entire native module were covariant with `'static`. I'm uncertain if things like threads and are guaranteed to be cleaned up, which would make this. Then again, the design of the `Env` API currently assumes this, being `Copy`, so...🤷🏻 I'll let you make the call there!

RE: @Xaeroxe's commentary about "just" allocating once here: I don't see any significant objection to the suggestion. N.B., I'd adapt the suggestion to `OsString`, given the above. I doubt a single string allocation here would be significant for the vast majority of applications, since AIUI the module file name is only computed once, and then cached for the remainder of the native module's lifetime.

ErichDonGubler:
Apologies for taking so long to get back to this! Finding time for OSS contributions while finding a job has been a bit turbulent for me, but I think I can push this over the line now. :) Responding to feedback now.

crates/napi/src/env.rs
ErichDonGubler: @Brooooooklyn:

RE: using `str` instead of `CStr`: I believe that the correct “strong” type to use here would be `OsStr`, rather than `str`, actually. The `TODO` remaining in this PR that I've [called out separately](https://github.com/napi-rs/napi-rs/pull/1232#discussion_r91703249) acknowledges that there's no documented guarantee of UTF-8 contents. However, since we're checking for the `file://` protocol prefix first, I think `OsStr` would be an exact match for the semantics specified by [Node-API upstream's docs for `node_api_get_module_file_name`](https://nodejs.org/docs/latest-v18.x/api/n-api.html#n_api_node_api_get_module_file_name) (emphasis mine):

> A URL containing the **absolute path of the location from which the add-on was loaded. For a file on the local file system it will start with `file://`.** The string is null-terminated and owned by `env` and must thus not be modified or freed.

ErichDonGubler: Ah, but I see that `OsStr` has a complication...the actual encoding accepted is platform-dependent, and constructors for it only exist for the `std::ffi::{unix,wasm}` modules. 😝

I think relying on UTF-8 paths _as an implementation_ is definitely acceptable as an MVP, which is then immediately convertible to `OsStr`. Not all valid system paths are UTF-8, but it should cover most folks' use cases until something better is found for parsing and constructing, i.e., [`os_str_bytes`](https://docs.rs/os_str_bytes/).

ErichDonGubler:
One other issue: I'm no longer an employee of @NZXTCorp, so I don't have access to the branch governing this PR. @Xaeroxe, did you want to push this over the line, or give me access?

crates/napi/src/env.rs
ErichDonGubler: RE: `Option<_>` and the proposed `get_module_file_name{,_raw}` split: I think returning `Result<$str_type, E>` instead, where `E` can give access to the original `CStr{,ing}`, would obviate the need for a split API, and be more idiomatic overall. If a user doesn't care what the failure is, they can call [`Result::ok`](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok) to immediately convert into an `Option<_>`.

Xaeroxe:
@ErichDonGubler Please copy the branch to your own fork and then re-open this PR. I cannot grant you permission to access that fork, nor do I have time to finish this myself.

Brooooooklyn:
Closed in https://github.com/napi-rs/napi-rs/pull/1783/commits/a09c579d67710d38f0a734505f4e278f82761e70

PR #1233, opened 2022-07-09
fix(napi): use Mutex rather than Atomic in ThreadSafeFunction

Brooooooklyn:


Brooooooklyn:
@SevInf please review 👀

PR #1234, opened 2022-07-09
fix(napi): use Mutex instead of Atomic in ThreadsafeFunction

Brooooooklyn:

### DIFF
--- a/crates/napi/src/threadsafe_function.rs
+++ b/crates/napi/src/threadsafe_function.rs
@@ -5,8 +5,8 @@ use std::ffi::CString;
 use std::marker::PhantomData;
 use std::os::raw::c_void;
 use std::ptr;
-use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
-use std::sync::Arc;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::{Arc, Mutex};
 
 use crate::bindgen_runtime::ToNapiValue;
 use crate::{check_status, sys, Env, Error, JsError, Result, Status};
@@ -146,20 +146,27 @@ type_level_enum! {
 /// ```
 pub struct ThreadsafeFunction<T: 'static, ES: ErrorStrategy::T = ErrorStrategy::CalleeHandled> {
   raw_tsfn: sys::napi_threadsafe_function,
-  aborted: Arc<AtomicBool>,
+  aborted: Arc<Mutex<bool>>,
   ref_count: Arc<AtomicUsize>,
   _phantom: PhantomData<(T, ES)>,
 }
 
 impl<T: 'static, ES: ErrorStrategy::T> Clone for ThreadsafeFunction<T, ES> {
   fn clone(&self) -> Self {
-    if !self.aborted.load(Ordering::Acquire) {
+    let is_aborted = self.aborted.lock().unwrap();
+    if !*is_aborted {
       let acquire_status = unsafe { sys::napi_acquire_threadsafe_function(self.raw_tsfn) };
       debug_assert!(
         acquire_status == sys::Status::napi_ok,
         "Acquire threadsafe function failed in clone"
       );
+    } else {
+      panic!("ThreadsafeFunction was aborted, can not clone it");
     }
+
+    self.ref_count.fetch_add(1, Ordering::AcqRel);
+
+    drop(is_aborted);
 
     Self {
       raw_tsfn: self.raw_tsfn,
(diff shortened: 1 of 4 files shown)


crates/napi/src/threadsafe_function.rs
```diff
@@ -231,39 +234,48 @@ impl<T: 'static, ES: ErrorStrategy::T> ThreadsafeFunction<T, ES> {
...
-    if self.aborted.load(Ordering::Acquire) {
+    let is_aborted = self.aborted.lock().unwrap();
+    if *is_aborted {
       return Err(Error::new(
         Status::Closing,
         "Can not ref, Thread safe function already aborted".to_string(),
       ));
     }
+    drop(is_aborted);
+    self.ref_count.fetch_add(1, Ordering::AcqRel);
```
SevInf, line 245: I don't quite understand ref counter usage in `refer`/`unref`. Original napi calls are not referenced counted and idempotent. And they don't control memory allocation, `release`/`acquire` do. So, while this fixes QueryEngine problem, doesn't it also introduce leak in other direction? For example:

```rust
let mut tsfn: ThreadsafeFunction<String> =
    func.create_threadsafe_function(0,...)

  tsfn.refer(&env)?;
``` 

This `refer` call is perfectly safe when using underlying APIs, albeit a little bit useless, but now it will cause function to leak.

crates/napi/src/threadsafe_function.rs
```diff
@@ -211,17 +215,16 @@ impl<T: 'static, ES: ErrorStrategy::T> ThreadsafeFunction<T, ES> {
         initial_thread_count,
```
SevInf, line 215: This might've been an issue already before, but I think there is no safe way to release a function, created with `inital_thread_count > 1`. The only place where release could happens is `drop`, and even there, `release` will be skipped if `ref_count` (which is initialized with `initial_thread_count`) is greater than 1. 

Not sure how to fix it though. Maybe it makes sense to remove the ability to set `initial_thread_count` from public API altogether and always use 1 internally? You'll still need to clone `ThreadsafeFunction` to use it from multiple threads, so `initial_thread_count > 1` does not make much sense in Rust world anyway.

crates/napi/src/threadsafe_function.rs
Brooooooklyn: The `refer/unref` API should not exist at all, it should be exposed as `clone/drop` in `Rust` world. 🤔 Maybe we should remove it in next major release.

> This refer call is perfectly safe when using underlying APIs, albeit a little bit useless, but now it will cause function to leak.

Nothing leak after this PR, the `Arc` will be dropped in the `fn drop` in `ThreadsafeFunction`

crates/napi/src/threadsafe_function.rs
Brooooooklyn: Yes, `initial_thread_count` is not public, and it's always `1`

crates/napi/src/threadsafe_function.rs
SevInf: Sorry, i think you are mixing up `ref`/`unref` and `acquire`/`release` API. 
`refer`/`unref` do not have anything to do with threadsafe function reference counting, despite what their name might suggest. They solely exist for letting or forbidding llibuv event loop completion without function cleanup. No matter how many times you call `napi_unref_threadsafe_function` it won't release the function and no matter how many times you call `napi_ref_threadsafe_function` it won't prevent function from being released. Moreover, no matter how many times you'd call  `napi_ref_threadsafe_function`, single `napi_unref_threadsafe_function` call will undo everything.  

So that's why adding reference counter in Rust wrapper does not make sense to me — this is not how N-API calls you are wrapping work.


Now let's walk through my example:
1. New function is created, internal reference counter is at 1.
2. We call `refer`. Original C-API does not do anything since function is already referenced. Rust wrapper on the other hand increments internal counter by 1, it is now at 2.
3. We drop the function. Rust internal ref counter is at 2. We skip release call, decrement the counter and finish the drop. Now neither the function is released, nor we have any pointer to the function. Function and it's context is leaked.

SevInf: This the relevant piece from [NodeApi docs](https://nodejs.org/api/n-api.html#deciding-whether-to-keep-the-process-running) I mentioned in my original report:

>Neither does napi_unref_threadsafe_function mark the thread-safe functions as able to be destroyed nor does napi_ref_threadsafe_function prevent it from being destroyed.

PR #1251, opened 2022-07-28
feat: use `CARGO_TARGET_DIR` if set

amrbashir:


cli/src/build.ts
```diff
@@ -409,7 +409,7 @@ export class BuildCommand extends Command {
       }
     }
 
-    const targetRootDir = await findUp(cwd)
+    const targetRootDir = process.env.CARGO_TARGET_DIR || await findUp(cwd)
```
amrbashir: `||` is desired here to catch the case when the env var is an empty string, should I disable eslint for this line?

Brooooooklyn:
@amrbashir lint failed

amrbashir:
@Brooooooklyn  I explained earlier that `||` is desired here instead of `??` to eliminate falsy values which is useful for situations when `CARGO_TARGET_DIR` is defined but is an empty string. 

I could use `??` which is enough for me but it won't catch the forementioned case or I could add an ignore lint for this line comment, what do you want me to do?

cli/src/build.ts
Brooooooklyn: Ah sorry, I missed this comment. I think to add eslint ignore here is reasonable.

Thanks for your PR

PR #1263, opened 2022-08-11
feat(napi): support rust array to js array

xhuz:
Support rust array to js array as follows:

```rust
let arr = [1u32; 10];

Array::from_vec(&env, arr);

Array::from_ref_vec(&env, &arr);

```


Brooooooklyn:
Thanks for the contirbution. Could you also add some tests in https://github.com/napi-rs/napi-rs/tree/main/examples/napi ?

PR #1281, opened 2022-08-20
Fix segfault when `ThreadsafeFunction`'s callback closure captures data

### LINKED ISSUE #1279: Segfault when using a value moved into `ThreadsafeFunction`'s callback
I'm not quite sure if this is my mistake but when moving a value into the callback of a ``ThreadsafeFunction`` this value will be NULL at execution causing the program to segfault. I also didn't find anything in the documentation stating that you can't move values into the callback closure. But correct me if I'm wrong.

Minimal example for ``napi-rs@2.8.0``:

```rs
#![deny(clippy::all)]

#[macro_use]
extern crate napi_derive;

use napi::threadsafe_function::{ThreadsafeFunction, ThreadsafeFunctionCallMode};
use napi::{JsFunction, JsString};

#[napi]
fn segfault(func: JsFunction) -> napi::Result<()> {
  let str = "test";
  let tsfn: ThreadsafeFunction<()> = func
    .create_threadsafe_function(0, move |_| {
      println!("{}", str); // str is NULL at this point
      Ok(Vec::<JsString>::new())
    })
    .unwrap();

  tsfn.call(Ok(()), ThreadsafeFunctionCallMode::NonBlocking);

  Ok(())
}
```

And the javascript side:
```js
const {segfault} = require("./index");

segfault(() => {});
```

messense:
Fixes #1279

Culprit commit: https://github.com/napi-rs/napi-rs/commit/4dfc770c2ae9aa4e27463570f062ce84ed5147f9

Valgrind report

```
==566183== HEAP SUMMARY:
==566183==     in use at exit: 90,986 bytes in 373 blocks
==566183==   total heap usage: 17,343 allocs, 16,970 frees, 17,878,110 bytes allocated
==566183==
==566183== LEAK SUMMARY:
==566183==    definitely lost: 0 bytes in 0 blocks
==566183==    indirectly lost: 0 bytes in 0 blocks
==566183==      possibly lost: 28,426 bytes in 161 blocks
==566183==    still reachable: 62,560 bytes in 212 blocks
==566183==                       of which reachable via heuristic:
==566183==                         multipleinheritance: 48 bytes in 1 blocks
==566183==         suppressed: 0 bytes in 0 blocks
==566183== Rerun with --leak-check=full to see details of leaked memory
==566183==
==566183== For lists of detected and suppressed errors, rerun with: -s
==566183== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```


crates/napi/src/threadsafe_function.rs
```diff
@@ -355,8 +355,8 @@ unsafe extern "C" fn thread_finalize_cb<T: 'static, V: ToNapiValue, R>(
   R: 'static + Send + FnMut(ThreadSafeCallContext<T>) -> Result<Vec<V>>,
 {
   // cleanup
-  drop(unsafe { Box::<R>::from_raw(finalize_data.cast()) });
-  let aborted = unsafe { Arc::<Mutex<bool>>::from_raw(finalize_hint.cast()) };
+  drop(unsafe { Box::<R>::from_raw(finalize_hint.cast()) });
```
messense, line 358: `finalize_hint` is the `context` passed into `napi_create_threadsafe_function `, see https://github.com/nodejs/node/blob/705fefd30d68e70b0b6aff07567ee0afb8324f89/src/node_api.cc#L376

Brooooooklyn: Great to hear 👍🏿

Issue #1292, opened 2022-08-31
Dependency syn needs a bumped minimum version


mat-if:
It appears that starting in version 2.7, napi started using some feature of `syn` that was introduced in version 1.0.61. The version given in the `Cargo.toml` for `napi-derive` and `napi-derive-backend` only specifies a minimum of `1.0`. When trying to upgrade to a newer napi version, I was running into compilation errors until I manually updated `syn` with `cargo upgrade -p syn`. 

`1.0.61` was published on 2021-03-05: https://lib.rs/crates/syn/versions

PR #1293, opened 2022-08-31
pin minimum usable version of dependency syn

mat-if:
Fixes https://github.com/napi-rs/napi-rs/issues/1292

This format still allows the same range of versions greater than 1.0.61 according to the docs:  https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

1 is equal to >=1, <2
1.0 is equal to >=1, <2
1.0.61 is equal to >=1.0.61, <2


Brooooooklyn:
Good catch, thank you 🚀

PR #1329, opened 2022-10-01
Switch from `toml` to `@ltd/j-toml`

### LINKED ISSUE #1328: cli fails with cargo workspace-inheritance
`Type Error: Could not parse the Cargo.toml`

napi commands will fail if using workspace inheritance within your toml file. 

```toml
[package]
edition = "2021"
version.workspace = true 
```

messense:
https://www.npmjs.com/package/@ltd/j-toml, it's the only [TOML 1.0 compliant parser](https://github.com/toml-lang/toml/wiki#v100-compliant) for JavaScript at the moment.

Fixes #1328

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -40,6 +40,7 @@
     "url": "https://github.com/napi-rs/napi-rs/issues"
   },
   "devDependencies": {
+    "@ltd/j-toml": "^1.35.2",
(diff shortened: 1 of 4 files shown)


messense:
`@ltd/j-toml` requires Node.js 14+.

messense:
IMO #1330 might be a better solution.

PR #1339, opened 2022-10-07
Fixes #1338 use after free with async, and fixes #1340

### LINKED ISSUE #1338: Asynchronous functions are unsound in the presence of the garbage collector
# Actual Output

1. The infinite loop prints a few times
2. The garbage collector runs (forced in this case, randomly in real production code)
3. The value is dropped ⚠️ 
4. The loop continues execution, *reading from the dropped value* ⚠️ ⚠️ ⚠️ 

```
Awesome(100) has been created
Awesome(100) is idling...
Awesome(100) is idling...
Awesome(100) is idling...
Awesome(100) is idling...
GC request (true)
GC request (false)
Awesome(99) has been dropped
Awesome(48) is idling...
Awesome(48) is idling...
Awesome(48) is idling...
```

# Expected

I expected the call to `runForever` to keep a reference to the `awesome` instance alive until the function finished execution.

# Steps to reproduce

**lib.rs**

```rust
use napi_derive::napi;
use std::time::Duration;
use tokio::time;

#[napi]
#[derive(Debug)]
pub struct Awesome(u8);

#[napi]
impl Awesome {
    #[napi(constructor)]
    pub fn new() -> Self {
        let this = Self(100);
        eprintln!("{this:?} has been created");
        this
    }

    #[napi]
    pub async fn run_forever(&self) {
        let mut i = time::interval(Duration::from_secs(1));
        loop {
            i.tick().await;
            eprintln!("{self:?} is idling...");
        }
    }
}

impl Drop for Awesome {
    fn drop(&mut self) {
        self.0 -= 1;
        eprintln!("{self:?} has been dropped")
    }
}
```

**Cargo.toml**

```toml
[package]
edition = "2021"
name = "napi-gc"
version = "0.0.0"

[lib]
crate-type = ["cdylib"]

[dependencies]
napi = { version = "2.10.0", default-features = false, features = ["napi8", "tokio_rt"] }
napi-derive = "2.9.1"
tokio = { version = "*", features = ["full"] }

[build-dependencies]
napi-build = "2.0.1"

[profile.release]
lto = true
```

**run.js**

```ts
const { Awesome } = require('./');

const awesome = new Awesome();

awesome.runForever().then(() => {
  console.log("runForever has exited");
});

setTimeout(() => {
  global.gc(true);
  console.log("GC request (true)");

  global.gc(false);
  console.log("GC request (false)");

  setTimeout(() => {
    process.exit();
  }, 3000);
}, 3000);
```

**package.json**

```json
{
  "name": "napi-gc",
  "version": "0.0.0",
  "main": "index.js",
  "types": "index.d.ts",
  "napi": {
    "name": "napi-gc",
    "triples": {
      "defaults": false,
      "additional": [
        "aarch64-apple-darwin"
      ]
    }
  },
  "license": "MIT",
  "devDependencies": {
    "@napi-rs/cli": "^2.12.0",
    "ava": "^4.3.3"
  },
  "ava": {
    "timeout": "3m"
  },
  "engines": {
    "node": ">= 10"
  },
  "scripts": {
    "x": "node --expose-gc run",

    "artifacts": "napi artifacts",
    "build": "napi build --platform --release",
    "build:debug": "napi build --platform",
    "prepublishOnly": "napi prepublish -t npm",
    "test": "ava",
    "version": "napi version"
  },
  "packageManager": "yarn@3.2.4"
}
```

Running `yarn run x`.

### LINKED ISSUE #1340: Asynchronous functions can mutably access the same memory simultaneously
Example project: [napi-mut-twice.zip](https://github.com/napi-rs/napi-rs/files/9750099/napi-mut-twice.zip)

This is in a similar vein to #1338. To run the example

```
yarn run build
```

This runs correctly, nothing is wrong.
```
yarn run x
```

This mutably borrows data twice, something is wrong.
```
yarn run x2
```

Expected behavior: I'm not sure. More on this later.

Actual behavior: In `yarn run x` we can see the code behaving in a normal and intended way. In `x2` however, the `tokio` runtime has two mutable borrows to the same data, which results in undefined and unsound behavior. This is not allowed per [Chapter 4 section 2 of the Rust Book](https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#mutable-references).

So what should we do about this? I have some ideas with varying trade offs.

1. Forbid `&mut _` in `async` `#[napi]` methods.

	Pros: JavaScript never has to think about Rust borrow rules and restrictions
	Cons: Rust users can't use `&mut _` with `async` functions anymore.

2. Use [RefCell](https://doc.rust-lang.org/stable/std/cell/struct.RefCell.html) or something like it to dynamically check borrow rules.

	Pros: We can keep writing async functions with `&mut _` arguments.
	Cons: JavaScript users are now exposed to the restrictions of Rust and all that entails in a likely surprising way.

My personal preference is for option 1.

Xaeroxe:
Fixes #1338 

EDIT: Now also fixes #1340


Xaeroxe:
Fix based on https://nodejs.org/api/n-api.html#object-lifetime-management

Xaeroxe:
This bumps the MSRV to 1.63 so that the fix can make use of [`array::from_fn`](https://doc.rust-lang.org/stable/std/array/fn.from_fn.html)

crates/backend/src/codegen/fn.rs
```diff
@@ -35,7 +35,40 @@ impl TryToTokens for NapiFn {
...
+                unsafe { napi::bindgen_prelude::sys::napi_delete_reference(env, r) },
+                napi::bindgen_prelude::sys::Status::napi_ok,
+                "failed to delete napi ref"
+              );
+            }
+          }
+        }
+        unsafe impl Send for NapiRefContainer {}
+        unsafe impl Sync for NapiRefContainer {}
+        let _make_ref = |a: ::std::ptr::NonNull<napi::bindgen_prelude::sys::napi_value__>| {
```
Brooooooklyn: there are some types that can't be applied to `napi_create_reference`, like `String` and `Number` in JavaScript 🤔

Brooooooklyn: Don't know if it would be better if handle this case here https://github.com/napi-rs/napi-rs/blob/main/crates/backend/src/codegen/fn.rs#L219

Xaeroxe: Ah, perhaps a combination of the two would work! I can see the utility in moving the reference capture to the place you mention, though I think the reference delete needs to stay where it is.

Xaeroxe: Addressed in [`eafebd6`](https://github.com/napi-rs/napi-rs/pull/1339/commits/eafebd6199f547f26fedd0a84d6be9f7d14af5ff)

Xaeroxe:
@Brooooooklyn this is ready for review

Xaeroxe:
I added a commit to this PR which implements my first solution for #1340

Brooooooklyn:
@Xaeroxe thanks! I'll add some tests before I merge this

Brooooooklyn:
@Xaeroxe I decide to allow mut ref with `unsafe` mark

Xaeroxe:
Sure, that seems good to me

PR #1348, opened 2022-10-21
fix(napi): BigInt::get_u64 lossless check

### LINKED ISSUE #1347: BigInt::get_u64 does not correctly return losslessness
`BigInt::get_u64` determines lossnessness by validating:

1. whether the value *has* a `sign_bit`
2. it consists of a single word

As such: `self.sign_bit && self.words.len() == 1`.

However, it is my belief that we should check whether the `BigInt` *does not have* a `sign_bit`

Wodann:
Fixes #1347.


Brooooooklyn (approved):
Thanks!

PR #1376, opened 2022-11-29
Bug fix: typed arrays ref shouldn't use offset.

nihohit:
Notice from the n-api docs that the data returned from `napi_get_typedarray_info` is already adjusted by the byte offset. https://nodejs.org/api/n-api.html#napi_get_typedarray_info

This means that when `as_ref`/`as_mut` apply the byte offset, the offset is in practice applied twice.
This wasn't caught in tests because no test tried to modify a typed array with a byte offset, and the test didn't us the typed array structs, only `JsTypedArray`. If you want, I can modify the rest of the functions in examples/napi-compt-mode/src/arraybuffers.rs and the matching tests, to test all typed arrays.

IMO the `byte_offset` field can be removed entirely from the struct, but I wanted to submit a minimal PR.


nihohit:
@Brooooooklyn, can you please release a version with this fix?

PR #1383, opened 2022-12-09
Better error message for `#[napi(catch_unwind)]`

simonvandel:
Before, the JS error would only have `Any { .. }`


Brooooooklyn (approved):
Thanks!

PR #1391, opened 2022-12-14
fix template to works with nix

### LINKED ISSUE #1335: Doesn't work in nixos
The template assumes that `ldd` is here:
```
 isMusl = readFileSync('/usr/bin/ldd', 'utf8').includes('musl')
```

It should do `which ldd` to find out its location and then it should work fine in nixos.

pepoviola:
Fixes #1335, checking the path of `ldd` before trying to read it.

Thanks!!


pepoviola:
thanks!!

Issue #1380, opened 2022-12-08
`napi_run_script` is not exported in `Env` type


F001:
Is it intended to not have this method or it is just unimplemented ?
Can I create a PR to add `run_script` method for `Env` type?
Thanks!

F001:
```rust
  pub fn run_script(&self, script: &str)->Result<JsObject> {
    let s = self.create_string(script)?;
    let mut raw_value = ptr::null_mut();
    check_status!(unsafe { sys::napi_run_script(self.0, s.raw(), &mut raw_value) })?;
    Ok(unsafe { JsObject::from_raw_unchecked(self.0, raw_value) })
  }
```

PR #1393, opened 2022-12-15
add "run_script" for "Env"

F001:
Fix https://github.com/napi-rs/napi-rs/issues/1380

### DIFF
--- a/crates/napi/src/env.rs
+++ b/crates/napi/src/env.rs
@@ -1020,6 +1020,13 @@ impl Env {
     check_status!(unsafe { sys::napi_close_handle_scope(self.0, handle_scope) })?;
     result
   }
+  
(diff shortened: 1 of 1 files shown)


crates/napi/src/env.rs
```diff
@@ -1020,6 +1020,13 @@ impl Env {
     check_status!(unsafe { sys::napi_close_handle_scope(self.0, handle_scope) })?;
     result
   }
+  
+  pub fn run_script(&self, script: &str)->Result<JsObject> {
```
Brooooooklyn:
```suggestion
  pub fn run_script<S: AsRef<str>>(&self, script: S)->Result<JsObject> {
```

Brooooooklyn (approved):
Thanks!

PR #1396, opened 2022-12-16
fix(napi): napi_create_async_work incorrect argument (napi-rs#1392)

patrickpilch:
The third argument to `napi_create_async_work` (async_resource_name) should be a `napi_value` that corresponds to a null-terminated utf-8 string.

This resolves errors running Next.js with GraalVM.

oracle/graal#5582
oracle/graal#5581


Brooooooklyn (approved):
Thanks!

PR #1397, opened 2022-12-16
Darwin universal architecture

skirsdeda:
**TLDR:** adds third 'universal' synthetic arch for macOS which combines x64/arm64 binaries into a universal binary and produces a single npm subdependency which is only tied to OS as opposed to OS/CPU combo.

### Motivation
I am using napi-rs to create native macOS extension for electron App. It was working fine when building electron bundle on x64 host for x64 target, but I am now switching to arm64 host while still building for x64 target. This doesn't work due to the way how npm resolves optional dependencies and there's no good way to install x64 optional dependency on arm64 host. So instead of hacking something to workaround this, I decided it would be better to have an option to build universal binaries in napi-rs. That way, I will be able to build for both x64 and universal targets on arm64.

### Implementation
- new triple 'universal-apple-darwin'
- new cli command 'universal' to combine x64/arm64 binaries
- index.js changes to first try universal before going for x64/arm64
- ci template changes as well as some unrelated fixes (typos and unnecessary stuff removed for darwin-arm64)

### Future considerations
Whenever nodejs becomes officially available for Windows on ARM, this could be extended to also support universal windows binaries.

And huge thanks for creating napi-rs, BTW 💯 It is perfect for me 😸

### DIFF
--- a/cli/src/__test__/parse-triple.spec.ts
+++ b/cli/src/__test__/parse-triple.spec.ts
@@ -113,6 +113,16 @@ const triples = [
       raw: 'armv7-linux-androideabi',
     },
   } as const,
+  {
+    name: 'universal-apple-darwin',
+    expected: {
+      abi: null,
+      arch: 'universal',
+      platform: 'darwin',
+      platformArchABI: 'darwin-universal',
+      raw: 'universal-apple-darwin',
+    },
+  } as const,
 ]
 
 for (const triple of triples) {
--- a/cli/src/artifacts.ts
+++ b/cli/src/artifacts.ts
@@ -60,6 +60,7 @@ export class ArtifactsCommand extends Command {
                 _binaryName,
               )}] is not matched with [${chalk.greenBright(binaryName)}], skip`,
             )
+            return
           }
           const dir = distDirs.find((dir) => dir.includes(platformArchABI))
           if (!dir && universalSourceBins.has(platformArchABI)) {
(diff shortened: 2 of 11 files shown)


CMCDragonkai:
To make use of this, can we just use the option `napi build --target universal-darwin-apple`?

Because I'm getting this atm:

```
  error: Error loading target specification: Could not find specification for target "universal-apple-darwin". Run `rustc --print target-list` for a list of built-in targets
```

What is the purpose of the new `universal` command?

CMCDragonkai:
Oh I think I understand, we have to build the x64 and arm64 separately...

Then we can try to combine them:

> new cli command 'universal' to combine x64/arm64 binaries

It's possible for an x64 macos to build the arm64 binary too right?

Brooooooklyn:
> It's possible for an x64 macos to build the arm64 binary too right?

Yes

CMCDragonkai:
Ok I've built 2 binaries.. But I've renamed them to what I want to use them for.

```
cmcdragonkai@matrix-mac-1 js-quic % ls prebuild/ 
quic-darwin-arm64.node	quic-darwin-x64.node
```

Now if I try to use `napi universal -d prebuild`. I just get an error about `config`.

Any docs about how this is supposed to be used? I see that the code seems to expect that the names of the binaries have to be specific. Can this command be made a bit more generic so I can just pass the binaries that I want combined?

skirsdeda:
@CMCDragonkai cli command was made to work with Github Actions. Dependence on source filenames is because that's how particular platform/arch binaries are looked up. If it doesn't fit your use-case, you can just use `lipo` on macOS directly.

PR #1413, opened 2022-12-28
test: Memory leak reproduction for futures

SevInf:
Hello!
When attempting to update to napi-rs 2.10.0, our tests caught a new [memory leak](https://github.com/prisma/prisma/pull/16258#issuecomment-1313793203) across all Node.JS version. I can confirm that the issue still exists on 2.10.3 as well. After some digging through, I think I narrowed it down to Future -> Promise conversion. This PR adds a reproduction to the memory test suite: if you run the script manually, you can see heap usage steadily climbing. And if you use low enough `--max_old_space_size` it will eventually crash with out-of-memory error. Hope it will help with fixing it — at the moment we had to pin all `napi-rs` packages to 2.9.0, but we'd like to unblock the update.


xiaopengli89:
I think it's related to #1403

PR #1432, opened 2023-01-11
Fix broken zig behaviors

### LINKED ISSUE #1430: Can’t use Zig “cross compilation” for native architecture
The Zig compiler is disabled when the target architecture/platform/abi matches the “native” environment, even if `--zig` is passed.

I’m interested in using Zig to create a `x86_64-unknown-linux-gnu` binary that supports a `glibc` version older than the `glibc` version on the platform I’m building on. I believe this is a supported use case of the `zig` linker, but it’s impossible to access this functionality through `napi build` unless you’re actually *cross*-compiling (from a different platform).

**To reproduce:** Run `napi build --platform --release --zig --target x86_64-unknown-linux-gnu` on an x86_64 linux machine with `libc` installed. Observe that zig is not invoked, and the binary is built with the system’s glibc headers rather than zig’s.

**Expected behavior:** zig can be opted-in by passing `--zig`

### LINKED ISSUE #1431: --zig-abi-suffix has no effect
napi-rs uses a zig `-target` of `x86_64-linux-gnu.2.17` when cross-compiling for `x86_64-unknown-linux-gnu` regardless of what value is passed to `--zig-abi-suffix` (and likewise for aarch64-unknown-linux-gnu)

This is because of a typo as far as I can tell; this expression:
https://github.com/napi-rs/napi-rs/blob/097c5d63db83200a46ddd2648a1b138d0f06e54e/cli/src/build.ts#L333-L336
always resolves to either `DEFAULT_GLIBC_TARGET` or `null`—because the `??` operator has higher precedence than the ternary. 

This bug was introduced in #1367.

controversial:
This PR includes 2 fixes:
1. fix #1430
  Zig wasn’t being activated when `--zig` is passed unless the build was also cross-platform. Under this PR, the `--zig` flag opts in to zig compilation regardless of platform.
3. fix #1431 
  `--zig-abi-suffix` was completely broken (its value did not change the suffix passed to zig). Under this PR, it works.

This PR also adds a debug message to show when zig is activated, so that (in debug mode) the user can tell whether their program is being compiled with `zig` or with the default compiler tools, as well as what target `zig` will compile for.

### DIFF
--- a/cli/src/build.ts
+++ b/cli/src/build.ts
@@ -309,11 +309,12 @@ export class BuildCommand extends Command {
     if (rustflags.length > 0) {
       additionalEnv['RUSTFLAGS'] = rustflags.join(' ')
     }
-    let isZigExisted = false
-    if (isCrossForLinux || isCrossForMacOS) {
+
+    let useZig = false
+    if (this.useZig || isCrossForLinux || isCrossForMacOS) {
       try {
         execSync('zig version')
-        isZigExisted = true
+        useZig = true
       } catch (e) {
         if (this.useZig) {
           throw new TypeError(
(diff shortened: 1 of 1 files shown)


controversial:
Updated to sign the commit!

cli/src/build.ts
```diff
@@ -329,15 +330,15 @@ export class BuildCommand extends Command {
       }
     }
 
-    if ((this.useZig || isCrossForLinux || isCrossForMacOS) && isZigExisted) {
+    if (useZig) {
```
controversial, line 333: Note that `useZig = true` implies all of the previous conditional (it can only be set true if `(this.useZig || isCrossForLinux || isCrossForMacOS)` and zig existed)

PR #1475, opened 2023-02-05
feat(napi): convert ToNapiValue tuple to variadic tsfn

h-a-n-a:
1. Implement Either like macro
2. Correct the TypeScript types generation
3. Fix a case where a single tuple may be generated as `any` for TypeScript declaration


crates/backend/src/typegen.rs
```diff
@@ -354,19 +409,25 @@ pub fn ty_to_ts_type(ty: &Type, is_return_ty: bool, is_struct_field: bool) -> (S
...
-    Type::Group(g) => ty_to_ts_type(&g.elem, is_return_ty, is_struct_field),
+    Type::Group(g) => ty_to_ts_type(&g.elem, is_return_ty, is_struct_field, false),
     Type::Array(a) => {
-      let (element_type, is_optional) = ty_to_ts_type(&a.elem, is_return_ty, is_struct_field);
-      (format!("{}[]", element_type), is_optional)
+      let (element_type, is_optional, _) =
+        ty_to_ts_type(&a.elem, is_return_ty, is_struct_field, false);
+      (format!("{}[]", element_type), is_optional, false)
+    }
+    Type::Paren(p) => {
```
h-a-n-a, line 424: Add this to fix types of tuples with only one element that exists.

h-a-n-a:
@Brooooooklyn Ready for review, but CI seems failed as the syntax is not rustc 1.57.0 compatible

crates/backend/src/typegen.rs
Brooooooklyn: great

Brooooooklyn (approved):

PR #1489, opened 2023-02-16
fix(cli): always generate typedef file even native code never changes

### LINKED ISSUE #1488: Bindgen file only generates on Rust code changes
Expected Results: 
* Any time I run `napi build`, the bindgen JavaScript file & .d.ts file should generate

Current Results: 
* If my rust code hasn't changed since a previous run of `napi build`, I delete all generated assets, and run `napi build` again, only the `.node` file is generated. 

Upon further inspection, it appears that this is caused by the following code: 

https://github.com/napi-rs/napi-rs/blob/853f52ed1f1d96882a033bfb7f7fd0c4f63d66ce/cli/src/build.ts#L605-L609

Where `idents` is only populated on file change and: 

https://github.com/napi-rs/napi-rs/blob/853f52ed1f1d96882a033bfb7f7fd0c4f63d66ce/cli/src/build.ts#L816-L822

Where the ident check occurs. 

I think we need to be calculating all idents each run, and not just changed.

forehalo:
close #1488

### DIFF
--- a/cli/src/build.ts
+++ b/cli/src/build.ts
@@ -1,4 +1,5 @@
 import { execSync } from 'child_process'
+import { createHash } from 'crypto'
 import { existsSync, mkdirSync } from 'fs'
 import { tmpdir } from 'os'
 import { join, parse, sep } from 'path'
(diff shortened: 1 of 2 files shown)


cli/src/build.ts
```diff
@@ -494,6 +468,39 @@ export class BuildCommand extends Command {
       debug(`Dylib name: ${chalk.greenBright(cargoArtifactName)}`)
     }
 
+    const intermediateTypeFile = join(
+      tmpdir(),
+      `${cargoArtifactName}.napi_type_def.tmp`,
```
Brooooooklyn: `${cargoArtifactName}-${sha256(process.cwd()).substring(0, 7)}.napi_type_def.tmp` to prevent naming conflict?

forehalo: cool

Brooooooklyn (approved):

Issue #1482, opened 2023-02-08
Create API duplicating the capabilities of the CLI.


TheLarkInn:
We are looking to introduce Rust into our monorepo management tool Rush. I have looked at the napi-rs/package-template and noticed that the scripts are very noisy. 

In a monorepo we depend on abstracting away that noise to make it easy for users to iterate. I would like to request that napi-rs introduce a NodeJS API that we can use programmatically to pass in configuration, outputpaths, destDir, and other behaviors found in the CLI. 

Today I'm forced to use `child_process.spawn()` to abstract away these complexities as seen in this PR:  https://github.com/microsoft/rushstack/pull/3961 

(specifically here https://github.com/microsoft/rushstack/pull/3961/files#diff-590cba6f6874b67951ecf78cf7ee4907cd6a9cd537a29cfd7ee71fbd461a4f6cR29-R58 )

forehalo:
@TheLarkInn

The api version of napi cli is available now in `@napi-rs/cli` prerelease phase. You can have a taste by `npm install @napi-rs/cli@alpha`.

We will ship it asap when we think it's stable enough and get all docs work done.

quick example:

```ts
import { NapiCli } from '@napi-rs/cli'

// more detail at https://github.com/napi-rs/napi-rs/tree/main/cli#readme
new NapiCli().build({
  // options
})
```

Brooooooklyn:
Close in https://github.com/napi-rs/napi-rs/pull/1492

Issue #1189, opened 2022-05-18
Limit generation of platform detection code to selected platforms


benwis:
Hi! I'm feeding napi-rs generated packages into esbuild for use with Remix, but it chokes on the generated index.js file due to the require() of platform .node files that don't exist.

Would it be possible to add a flag or option to limit the platform detection code to selected build targets?

Issue #1188, opened 2022-05-15
issues with `--pipe` option


kyranet:
I'm defining the following script in package.json, and with generated files being deleted:

```json
{
	"scripts": {
		"build": "napi build --platform --release --js index.cjs --pipe \"prettier -w\""
	}
}
```

And when running `yarn build`, prettier will fail to format `index.d.ts`, resulting on an unformatted file, but the JS output is fine:

![image](https://user-images.githubusercontent.com/24852502/168495887-066a123e-b68a-42ae-b93a-86530884f355.png)

Running it again will result on Prettier to format the file correctly... often. ~~I haven't been able to reproduce it again, but it once formatted the file and then overwrote with the unformatted code, so this might be caused by a race-condition?~~ Turns out, the built code is written *always* after the piped command finishes writing, which causes the command to output unformatted code. This can be reproduced by modifying the source code or deleting target to force a re-build.

---

Something interesting I found. If I delete both generated files, and then I run `yarn build`, the files won't auto-regenerate, and will still call the script, this causes the script to fail as shown in the screenshot below:

![image](https://user-images.githubusercontent.com/24852502/168495813-256aaacb-8c28-48cb-afb2-ae5f9ab7e8fa.png)

## Additional information

- `OS`: Windows 11 Pro (Version 21H2, OS Build 22000.675).
- `Rustc`: 1.60.0 (7737e0b5c 2022-04-04)
- `Node.js`: 18.1.0
- `@napi-rs/cli`: 2.9.0
- `prettier`: 2.6.2

Nytelife26:
This works for me on Ubuntu. Although, I should note that it seems unusual to have a `--pipe` option at all, and my recommendation would be to remove it altogether in favour of standard operating system pipes if possible - or, if not, at least have the command output the generated filenames so a script can be formed around it.

Brooooooklyn:
Closed in https://github.com/napi-rs/napi-rs/pull/1492

Issue #1170, opened 2022-05-07
Feature request: add custom target-dir support


0xDing:
Cargo can override the default target dir location by setting target-dir in config.toml, while napi-cli assumes that target-dir is ` . /target ` which leads to a `Type Error: No target dir found error.`

P.s. It makes sense to change the target dir in a large monorepo that contains both Typescript and Rust code

ref: https://github.com/napi-rs/napi-rs/blob/79d7453713edc26bb283f1b63c5314a06163f323/cli/src/build.ts#L517

Thanks

Brooooooklyn:
Closed in https://github.com/napi-rs/napi-rs/pull/1492

PR #1492, opened 2023-02-22
feat: rewrite napi-cli

forehalo:
Although the rust rewriting has been done for a while, we have a lot of concerns on releasing it.

As we received a requirement about programmatic cli #1482, so we think it's time to bring things up today.

preparation:
- [x] borrow things from rust cli
- [x] esm/cjs scaffold

features: 
- [x] napi new  
- [x] napi build
- [x] napi watch
- [x] napi xxx

Aimed to fix all known issues with current cli as well:
- [x] #988 
- [x] #407
- [x] #1061
- [x] #1170 
- [x] #1188 
- [x] #1482

Not included:
- [ ] #1045
- [ ] #1189


forehalo:
# Breaking changes:

## Node > 16

We drop the support of outdated LTS versions of node like v12 and v14. @napi-rs/cli now requires your node version to be greater then 16.

##ESM/CJS supports

```js
// ESM
import { NapiCli } from '@napi-rs/cli'

// CJS
const { NapiCli } = require('@napi-rs/cli')
```

## Programmatical API

```js
import { NapiCli } from '@napi-rs/cli'

new NapiCli().build({ ... })
```

## CLI option naming convention

| Before 3.0    | 3.0                          |
| ------------- | ---------------------------- |
| --zig-\*      | --cross-compile,-x           |
| --config,-c   | --package-json-path          |
| --cargo-cwd   | --manifest-path=./Cargo.toml |
| --cargo-flags | -- --any-pass-through-flags  |

Issue #1514, opened 2023-03-13
Using ThreadsafeFunction after refreshing the page in Electron will crash


HotQ:
### Recurrence path
- create a `threadsafe_function::ThreadsafeFunction` in electron renderer process,
- refresh page `(control+R)`,
- `ThreadsafeFunction::call` will cause crash with **bad access**.

Electron version: 19

### Root Cause
When renderer refreshes page, Eletron will reset the Node environment, and **libuv will not be pending for ThreadsafeFunction**; 
ThreadSafeFunction inside Node will deleted by `CleanupHook`, but `napi_rs` won't stop users from accessing it.

[Node 10](https://github.com/nodejs/node/blob/19f3e1702e9e732462fcf83939b99d3bd0587ed2/src/node_api.cc#L918)
[Node master](https://github.com/nodejs/node/blob/1ef4faa29bd5ef91462ce2dd1c3ec3e463d980b9/src/node_api.cc#L206)

BTW:  There are other possible crash paths with node api after refreshing Eletron's page, but AsyncTask works fine here :)

PR #1515, opened 2023-03-13
fix(napi):  prevent access to tsfn-raw after env's destroyed(#1514)

HotQ:
Because the life cycle of tsfn-raw is bound to node env, so we can add tsfn-raw to cleanup hook as well.

### DIFF
--- a/crates/napi/src/threadsafe_function.rs
+++ b/crates/napi/src/threadsafe_function.rs
@@ -4,8 +4,9 @@ use std::convert::Into;
 use std::ffi::CString;
 use std::marker::PhantomData;
 use std::os::raw::c_void;
-use std::ptr;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::pin::Pin;
+use std::ptr::{self, null_mut};
+use std::sync::atomic::{AtomicBool, AtomicPtr, Ordering};
 use std::sync::{Arc, RwLock};
 
 use crate::bindgen_runtime::{FromNapiValue, ToNapiValue, TypeName, ValidateNapiValue};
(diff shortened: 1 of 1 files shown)


crates/napi/src/status.rs
```diff
@@ -30,6 +30,8 @@ pub enum Status {
   DetachableArraybufferExpected,
   WouldDeadlock,
   NoExternalBuffersAllowed,
+  /// The napi_env passed into the call was not valid
+  EnvInvalid,
```
Brooooooklyn: Don't need an additional error code here, these statuses are aligned to https://nodejs.org/api/n-api.html#napi_status

crates/napi/src/threadsafe_function.rs
```diff
@@ -97,23 +98,83 @@ type_level_enum! {
 }
 
 struct ThreadsafeFunctionHandle {
-  raw: sys::napi_threadsafe_function,
+  raw: AtomicPtr<sys::napi_threadsafe_function__>,
   aborted: RwLock<bool>,
+  valid: AtomicBool,
```
Brooooooklyn: Don't need `valid` status here, reuse `aborted` is fine.

crates/napi/src/threadsafe_function.rs
```diff
@@ -97,23 +98,83 @@ type_level_enum! {
...
   referred: AtomicBool,
 }
 
 unsafe impl Send for ThreadsafeFunctionHandle {}
 unsafe impl Sync for ThreadsafeFunctionHandle {}
 
+impl ThreadsafeFunctionHandle {
+  // called when the env is being destroyed
+  // then we can't call napi_****_threadsafe_function any more
+  unsafe extern "C" fn clearup(self_ptr: *mut c_void) {
```
Brooooooklyn: move it out of the `impl` block

crates/napi/src/threadsafe_function.rs
HotQ: got it

crates/napi/src/threadsafe_function.rs
HotQ: Is it because of mangling or something?

Brooooooklyn: This is mainly for consistency with the rest of the project where `extern "C" fn` is used. Other places that don't put it in `impl`...

HotQ:
forget about cleanup hook, we can only use `thread_finalize_cb `to achieve this.

crates/napi/src/threadsafe_function.rs
```diff
@@ -307,20 +337,19 @@ impl<T: 'static, ES: ErrorStrategy::T> ThreadsafeFunction<T, ES> {
         async_resource_name,
         max_queue_size,
         1,
-        ptr::null_mut(),
+        Arc::into_raw(Pin::into_inner(handle.clone()))
+          .cast_mut()
```
Brooooooklyn: This API was added in `1.65.0`: https://doc.rust-lang.org/std/primitive.pointer.html#method.cast_mut
It breaks the MSRV

Brooooooklyn:
@HotQ have you tested this branch with Electron yet? Does this patch fix your problem?

Brooooooklyn:
@HotQ let's deal with the worker thread case in another PR, thanks!

Brooooooklyn:
@HotQ IMO, in this case: 

```rust
// got lock_guard if not aborted
if let Some(_guard) = tsfn.lock(){
    tsfn.call(Ok(()), NonBlocking)
}

pub fn lock(&self) -> Option<RwLockReadGuard<bool>> {
  let aborted_guard = self
    .handle
    .aborted
    .read()
    .expect("Threadsafe Function aborted lock failed");
  (!*aborted_guard).then_some(aborted_guard)
}
```

You even don't need to `guard` the `aborted` before calling. The `status` returned is enough to determine if it is successful or not.

HotQ:
tsfn's call is currently unprotected, it means cleanup and finalize will happen between will  judging whether abortedand tsfn being called——And that will cause a crash(the tsfn has been deleted inside nodejs will be dereferenced).
so I need a guard to prevent the main thread's finalizing and the worker thread's access to tsfn from happening one after another.

HotQ:
Or 
```rust
pub fn call(&self, value: Result<T>, mode: ThreadsafeFunctionCallMode) -> Status {
  let aborted_guard = self
      .handle
      .aborted
      .read()
      .expect("Threadsafe Function aborted lock failed");
  if !*aborted_guard {
    unsafe { sys::napi_call_threadsafe_function( ... ).into() }
  }else{
    Status::Cancelled
  }
}
`

PR #1525, opened 2023-03-18
fix(napi): use weak arc for passing `thread_finalize_data`

meowtec:
If `thread_finalize_data` uses Arc::clone(), the reference count will not decrease to 0 after `ThreadsafeFunction` dropped, which will also cause node process to not exit.

```rs
pub fn function(callback: JsFunction) -> Result<()>  {
  let function = callback
      .create_threadsafe_function(0, ...).unwrap();

  std::thread::spawn(move || {
    function.call(..., ThreadsafeFunctionCallMode::Blocking);
  });
  Ok(())
}
```

```js
// this progress won't exit automatically
function((() => {})
```

The problem was introduced in #1515

Additional:

`impl Send + Sync` seems to be unnecessary anymore.

BTW I also remove `Pin` because I don't well know it but it seems to work well after been removed. If I have some mistake please tell me.

### DIFF
--- a/crates/napi/src/threadsafe_function.rs
+++ b/crates/napi/src/threadsafe_function.rs
@@ -4,10 +4,9 @@ use std::convert::Into;
 use std::ffi::CString;
 use std::marker::PhantomData;
 use std::os::raw::c_void;
-use std::pin::Pin;
 use std::ptr::{self, null_mut};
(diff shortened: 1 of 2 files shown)


meowtec:
I tested with electron and found that #1515 might did not really fix #1514 - page still crash after reload.
Maybe need to check `aborted` before `napi_call_threadsafe_function`...

Issue #1378, opened 2022-12-06
Pass closure for Property setter/getter


MarkusJx:
Currently, you can already define a js function for an object from a closure like this:
```rust
this.set_named_property(
	"name",
	env.create_function_from_closure(
		"name",
		move |ctx: CallContext| -> napi::Result<JsUnknown> {
			method(&ctx, &variable_copied_from_above)
		},
	)?,
)?;
```

As I'd also like to copy variables from above into getters and setters, it would be great if I could do something like this:
```rust
this.define_properties(&[Property::new("name")?
	.with_getter(move |ctx: CallContext| -> napi::Result<JsUnknown> {
		getter_method(&ctx, &variable_copied_from_above)
	})
	.with_setter(move |ctx: CallContext| -> napi::Result<JsUnknown> {
		setter_method(&ctx, &variable_copied_from_above)
	})
	.with_property_attributes(PropertyAttributes::Enumerable)])?;
```

Just as a Background, I want to create objects with automatically generated getters and setters, thus, I need to pass the name of the getter and setter to the actual implementation of the getter/setter.

LeoLeSommer:
I made some researches about it while i was diving into the code for my own needs.

(I'm absolutely not an expert of rs-napi).

If you take a look at the implementation of create_function_from_closure, it use napi_sys::napi_create_function. Which is a simple wrapper over the C++ api.

```
    fn napi_create_function(
        env: napi_env,
        utf8name: *const c_char,
        length: usize,
        cb: napi_callback,
        data: *mut c_void,
        result: *mut napi_value,
      ) -> napi_status;
```

As you see, there is a data parameter, which is an any pointer to a value.

With this pointer, you can send every data you want, personally i'm just sending a reference to a string representing the name of the property i want to access.

The property descriptor has an analog way to send datas to the getter/setter function, if you take a look at Property::raw(), it looks like it:

```
  pub(crate) fn raw(&self) -> sys::napi_property_descriptor {
    sys::napi_property_descriptor {
      utf8name: self.name.as_ptr(),
      name: ptr::null_mut(),
      method: self.method,
      getter: self.getter,
      setter: self.setter,
      value: self.value,
      attributes: self.attrs.into(),
      data: ptr::null_mut(),
    }
  }
```

You can affect the properties to a js object with the following code (where properties is a vec of sys::napi_property_descriptor):

```
                    unsafe {
                        check_status!(napi_sys::napi_define_properties(
                            env.raw(),
[...]

PR #1526, opened 2023-03-19
feat(napi): property getter and setter with closure

victorteokw:
This PR tries to implement issue #1378


crates/napi/src/js_values/object_property.rs
```diff
@@ -77,11 +91,184 @@ impl Property {
...
+                &mut closures_data_pointer,
+              )
+            };
+            debug_assert!(
+              Status::from(status) == Status::Ok,
+              "napi_get_cb_info failed"
+            );
+            (raw_this, raw_args, closures_data_pointer)
+          };
+          let closures = unsafe { closures_data_pointer.cast::<PropertyClosures>().as_ref().expect("cannot cast") };
```
Brooooooklyn: use `Box::leak(Box::from_raw(closures_data_pointer))` to avoid `expect`

crates/napi/src/js_values/object_property.rs
```diff
@@ -77,11 +91,184 @@ impl Property {
...
```
Brooooooklyn: same here

victorteokw:
Hi @Brooooooklyn, Thanks for the review! 👏 

I'll update the code. By the way, we may need someway to detect these and register the cleanup from `define_properties`. It's all fine if a property with setter and getter closure is defined for a class prototype. If somebody defines this on plain objects, it may leak a little heap memory.

Brooooooklyn:
@victorteokw I've cleaned up your implementation and added some tests 😁 I'm going to merge this after all checks pass

victorteokw:
😻 Thanks! @Brooooooklyn

Brooooooklyn (approved):

PR #1531, opened 2023-03-21
fix(cli,napi-derive): re-export types from shared crate

Brooooooklyn:


ceifa:
All of my types are being duplicated, could it be a regression from this PR?
image

Brooooooklyn:
@ceifa have you upgraded the `napi-derive` and `@napi-rs/cli` as well?

PR #1533, opened 2023-03-21
Ensure that `napi_call_threadsafe_function` cannot be called after abort

meowtec:


Brooooooklyn:
@meowtec I'm wondering if we can add an integration test for that via electron... could you try it with this branch here?

meowtec:
> @meowtec I'm wondering if we can add an integration test for that via electron... could you try it with this branch here?
Yea.

meowtec:
The new electron test will always fail because of the runtime dropped problem: https://github.com/napi-rs/napi-rs/pull/1542
I will continue after #1542 get merged.

Issue #1537, opened 2023-03-22
Losing precision on serde_json::Value to JS object


ceifa:
Related https://github.com/ceifa/steamworks.js/issues/60

PR #1538, opened 2023-03-22
fix(napi): big numbers losing precision on serde_json::Value

ceifa:
Fixes https://github.com/napi-rs/napi-rs/issues/1537


examples/napi/src/serde.rs
```diff
@@ -30,3 +30,9 @@ fn get_package_json_name(package_json: PackageJson) -> String {
 fn test_serde_roundtrip(data: Value) -> Value {
   data
 }
+
+#[napi]
+fn test_serde_big_number_precision(number: String) -> Value {
+  let data = format!("{{\"number\":{number}}}");
```
Brooooooklyn:
```suggestion
  let data = format!("{{\"number\":{}}}", number);
```

crates/napi/src/bindgen_runtime/js_values/serde.rs
```diff
@@ -111,14 +113,33 @@ impl FromNapiValue for Map<String, Value> {
 
 impl ToNapiValue for Number {
   unsafe fn to_napi_value(env: sys::napi_env, n: Self) -> Result<sys::napi_value> {
+    #[cfg(feature = "napi6")]
+    let max_safe_int = (2f64.powi(53) - 1f64) as i64;
```
Brooooooklyn:
```suggestion
    const MAX_SAFE_INT: i64 = (2f64.powi(53) - 1f64) as i64;
```

ceifa:
@Brooooooklyn I'm not sure if I did the feature flag thing correctly, can you take a look?

crates/napi/src/bindgen_runtime/js_values/serde.rs
```diff
@@ -111,14 +113,33 @@ impl FromNapiValue for Map<String, Value> {
...
       let n = n.as_u64().unwrap();
       if n > u32::MAX as u64 {
-        todo!("impl BigInt")
+        #[cfg(feature = "napi6")]
+        {
+          return unsafe { BigInt::to_napi_value(env, BigInt::from(n)) };
+        }
+
+        #[cfg(not(feature = "napi6"))]
+        return Err(Error::new(
```
Brooooooklyn: create `String` instead?

crates/napi/src/bindgen_runtime/js_values/serde.rs
ceifa: This produces an error: cannot call non-const fn `std::f64::<impl f64>::powi` in constants calls in constants are limited to constant functions, tuple structs and tuple variants

I will substitute to the constant number and comment the expression

ceifa:
Clippy made this suggestion:
```rust
 if n > MAX_SAFE_INT || n < -MAX_SAFE_INT {
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use: `!(-MAX_SAFE_INT..=MAX_SAFE_INT).contains(&n)`
```

To be honest I think it's much more unreadable that way, what do you think @Brooooooklyn ?

Brooooooklyn:
I'd like to follow the `Clippy`

Brooooooklyn (approved):
Great work!

Issue #1356, opened 2022-10-31
Unable to run cargo test since napi=2.10.0


punkeel:
Running `cargo test` now fails with the following error:
`dyld[48222]: symbol not found in flat namespace (_napi_delete_reference)`

This worked fine in previous napi releases, up to (including) `napi=2.9.1` and `napi-derive=2.9.1`.

Environment:
- macOS 12.6
- cargo 1.64.0 (387270bc7 2022-09-16)

oyyd:
Initializing a new template with `napi=2.10.0`, and running a `cargo test` works fine in my Mac.

I guess some of your test code use napi and that's what unexpected. Because when `cargo test`, these symbols, e.g. `_napi_delete_reference`, are not linked into the execuatable. They are included when the rust lib(Node.js addon) is included and the code is running in a Node.js process.

punkeel:
Thanks for the link! I've tried the two workarounds with no luck (both on my project and on the template project).

Workaround 1:
```sh
$ rustup target list|grep installed
aarch64-apple-darwin (installed)

~/code/tmp/repro ❯ cat .cargo/config.toml
[target.aarch64-apple-darwin]
rustflags = ["-C", "link-args=-Wl,-undefined,dynamic_lookup"]
```
<details>
  <summary>logs</summary>

```
~/code/tmp/repro ❯ rm -rf target
~/code/tmp/repro ❯ cargo test -v
   Compiling proc-macro2 v1.0.47
   Compiling unicode-ident v1.0.5
   Compiling quote v1.0.21
   Compiling syn v1.0.103
   Compiling memchr v2.5.0
   Compiling once_cell v1.16.0
   Compiling regex-syntax v0.6.27
   Compiling napi-build v2.0.1
   Compiling convert_case v0.5.0
   Compiling napi-sys v2.2.2
[...]

Issue #1032, opened 2022-01-15
a couple questions about v2


bmacnaughton:
v2 looks amazing - thanks for the work. i have tried using it with our project and ran into a few issues that might be worth documenting in your blog. and i have a couple questions that i'm working on, but hoping you might know the answer.

- the `Property::new()` function no longer takes an env argument
- is there a way, using v2 syntax, to return `napi::Result<JsUnknown>` so that different types can be returned?
- when i submit our build to github actions, the link fails because it cannot find the `napi_` symbols. any ideas?
- there are a couple cases where my function just passed the `CallContext` to another function so it can decode arguments. it looks like using the v1 syntax is the best way to continue that.

bmacnaughton:
the link error occurs when `cargo test` is executed, and it occurs locally as well (i thought it was the build step but it's not). v1 executes `cargo test` without errors.

while most of our tests are javascript, there are also rust tests.

bmacnaughton:
i thought there might be a difference in the library path, but they're the same. the only differences are the metadata and extra-filename values (and the napi and napi-derive versions).

v1 command being executed for `cargo test`:
[...]

PR #1546, opened 2023-03-28
chore(napi): add noop feature in napi crate

gaoquanzero:
- Fix https://github.com/napi-rs/napi-rs/issues/1356

### DIFF
--- a/crates/napi/Cargo.toml
+++ b/crates/napi/Cargo.toml
@@ -20,6 +20,7 @@ all-features = true
 independent = true
 
 [features]
+noop = []
 async = ["tokio_rt"]
 compat-mode = []
 default = ["napi3", "compat-mode"]                                               # for most Node.js users
--- a/crates/napi/src/bindgen_runtime/module_register.rs
+++ b/crates/napi/src/bindgen_runtime/module_register.rs
@@ -200,7 +200,7 @@ pub fn get_class_constructor(js_name: &'static str) -> Option<sys::napi_ref> {
 }
 
 #[doc(hidden)]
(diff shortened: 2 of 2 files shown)


Brooooooklyn:
@gaoquanzero does this patch really fix the issue?

gaoquanzero:
> @gaoquanzero does this patch really fix the issue?

hi, @Brooooooklyn 
according to this comment [https://github.com/napi-rs/napi-rs/issues/1032#issuecomment-1230622852](https://github.com/napi-rs/napi-rs/issues/1032#issuecomment-1230622852), this only happens at `cargo test` where napi build is not available.
the `_napi_delete_reference` symbol is called here [https://github.com/napi-rs/napi-rs/blob/main/crates/napi/src/error.rs#L167](https://github.com/napi-rs/napi-rs/blob/main/crates/napi/src/error.rs#L167), but feature `noop` not exist in napi crate（it's declared in `napi-derive` crate）, so `#[cfg(not(feature = "noop"))]` makes no sense,  this patch just made it work



minimal reproduce
```rust
#[cfg(test)]
mod tests {
  use napi::bindgen_prelude::*;
  fn test() -> Result<i32> {
    Ok(0)
  }

  #[test]
  fn test_it() {
    test();
  }
}

```

```toml
[dependencies]
# Default enable napi4 feature, see https://nodejs.org/api/n-api.html#node-api-version-matrix
napi = { path ="../napi-rs-main/crates/napi", version = "2.12.0", default-features = false, features = ["napi4", "noop"] }
napi-derive = "2.12.2"

```

PR #1552, opened 2023-04-02
fix(napi): access violation in property getter/setter closure

MarkusJx:
@victorteokw already did the bulk of the work on this but forgot to access the closure pointers properly.
This caused an access violation to come up every time external data from outside the closure was being used.
This PR aims to fixing this issue by accessing the closures from the data pointer.

See also: #1526, #1378


Brooooooklyn (approved):

victorteokw:
@MarkusJx 👍 @Brooooooklyn 👍

PR #1592, opened 2023-05-17
feat(napi): Bigint deserialization

JanisGailis:
Hi!

This PR does the following:

* Handle little endianness;
* Make sure get_<u/i>128 methods don't panic when words.len() < 2;
* Make sure BigInt deserialization takes into account word length and sign;

For context:

I'm using `napi-rs` in a context where I need to deserialize random Typescript values into concrete Rust types that I can't control. So I get a `BigInt`, which I need to deserialize into u64. With the current `napi-rs` version, deserialization would panic if a `BigInt` is provided that fits into a single word. With this change the rust type `napi-rs` deserializes `BigInt` into depends on what the value actually holds in memory on node side. Also, node API docs outline that the words are little endian, but this wasn't handled in the current implementation, so added that too!

I'm happy to adjust the PR as needed if you have any concerns, as well as add a test case if you can shortly outline how to best do that!

Thanks a lot for your work on `napi-rs`!

### DIFF
--- a/crates/napi/src/js_values/bigint.rs
+++ b/crates/napi/src/js_values/bigint.rs
@@ -245,20 +245,49 @@ impl JsBigInt {
 
   pub fn get_i128(&mut self) -> Result<(i128, bool)> {
     let (signed, words) = self.get_words()?;
-    let len = words.len();
-    let i128_words: [i64; 2] = [words[0] as _, words[1] as _];
-    let mut val = unsafe { ptr::read(i128_words.as_ptr() as *const i128) };
+
+    let high_part = words.first().copied().unwrap_or(0).to_le_bytes();
+    let low_part = words.get(1).copied().unwrap_or(0).to_le_bytes();
+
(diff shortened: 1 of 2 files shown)


JanisGailis:
@Brooooooklyn Thanks for starting the pipelines! 

What's the process, should I fix the broken ones? Clippy and 1.57 compatibility I can do!

Not sure what's going on with the Benches and the failing armv7 test.

Brooooooklyn:
@JanisGailis armv7 test is flaky due to the emulator, it's passed now. And you can ignore the benchmark CI

JanisGailis:
Fixed clippy and msrv lints! :crossed_fingers:

JanisGailis:
@Brooooooklyn Is there anything else you'd like to be done to have  this merged? 

I think the FreeBSD test passed the first time around, I don't really see how the minor refactor when fixing the lints could've caused any issue in that test!

JanisGailis:
@Brooooooklyn Thank you!!

What's your release schedule?

PR #1598, opened 2023-05-26
feat(cli): detect Cargo `--profile` argument

lucasfernog:
I actually need this feature on the v2 CLI, do you think you could backport and publish it? I used patch-package [on the Tauri PR](https://github.com/tauri-apps/tauri/pull/6861) but it's.. ugly :smile: 

I wanted to push to v2 instead but I couldn't find how you are handling it, I guess you're doing it manually?

### DIFF
--- a/cli/src/api/build.ts
+++ b/cli/src/api/build.ts
@@ -460,12 +460,13 @@ class Builder {
       return
     }
 
-    const src = join(
-      this.targetDir,
-      this.target.triple,
-      this.options.release ? 'release' : 'debug',
-      srcName,
-    )
+    let outputFolder = this.options.release ? 'release' : 'debug'
+    if (this.options.cargoOptions?.includes('--profile')) {
+      const args = this.options.cargoOptions
+      const profileIndex = args.indexOf('--profile')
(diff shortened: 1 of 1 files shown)


lucasfernog:
Maybe at this point it's easier to add a `profile` argument instead of parsing `cargoOptions`? Let me know.

Brooooooklyn:
> Do you think you could backport and publish it? 

Yes sure, I'll backport it after this is merged

Brooooooklyn:
@lucasfernog backport to https://github.com/napi-rs/napi-rs/releases/tag/%40napi-rs%2Fcli%402.16.0

lucasfernog:
@Brooooooklyn the `this.profile ?? this.isRelease ? 'release' : 'debug'` logic is wrong. When `profile` is set you actually need to use that instead of `release | debug`. So the actual code would be something like:

```javascript
const targetDir = join(
    this.targetTripleDir,
    this.profile || (this.isRelease ? 'release' : 'debug'),
)
```

Brooooooklyn:
@lucasfernog fixed again

PR #1626, opened 2023-06-15
fix cli incorrect behaviors

forehalo:
- [x] module exports
- [x] engine requirements
- [x] target selector prompt


cli/src/api/templates/js-binding.ts
```diff
@@ -253,5 +253,6 @@ if (!nativeBinding) {
   throw new Error(\`Failed to load native binding\`)
 }
 
+module.exports = nativeBinding
```
Brooooooklyn: you should generate codes like this: https://github.com/napi-rs/napi-rs/blob/v2/cli/src/build.ts#L876

```js
const { foo, bar }  = nativeBinding
module.exports.foo = foo
module.exports.bar = bar
```

so that users can use `import { foo } from '@napi-rs/binding'` in esm projects

forehalo: any testcase? does `entries.forEach((k, v) => { module.exports[k] = v })` help?

PR #1631, opened 2023-06-20
fix cli watch build and remove unused `--esm` option

forehalo:


Brooooooklyn:
`src/api/new.ts(149,20): error TS2339: Property 'esm' does not exist on type 'Required<NewOptions>'.`

Brooooooklyn (approved):

PR #1645, opened 2023-07-07
fix(napi): CallContext closure lifetime issue

nawbc:
CallContext `'env` will cause a closure lifetime issue. See https://users.rust-lang.org/t/closure-lifetime-issue-argument-requires-that-1-must-outlive-static/96367  for more details.

NB: may cause breaking change.


Brooooooklyn:
This will increase call overhead for every function call

nawbc:
> This will increase call overhead for every function call

I will test performance changes later, and improve.

PR #1669, opened 2023-07-25
Integrate with emnapi

Brooooooklyn:
- [x] compile to `wasm32-wasi-preview1-threads`
- [x] generate wasm file and loader
- [ ] pass tests in example


crates/build/src/wasi.rs
```diff
@@ -0,0 +1,20 @@
+use std::env;
+
+pub fn setup() {
+  let link_dir = env::var("EMNAPI_LINK_DIR").expect("EMNAPI_LINK_DIR must be set");
+  println!("cargo:rustc-link-search={}", link_dir);
+  println!("cargo:rustc-link-lib=static=emnapi-basic-mt");
+  println!("cargo:rustc-link-arg=--initial-memory=16777216");
+  println!("cargo:rustc-link-arg=--export-dynamic");
+  println!("cargo:rustc-link-arg=--export-memory");
```
toyobayashi: wasi threads prefers `--import-memory`, see here: https://github.com/WebAssembly/wasi-sdk/issues/328.

also, `importObject.env.memory = new WebAssembly.Memory(...)` only works with `--import-memory`.

cli/src/api/load-wasi-template.ts
```diff
@@ -0,0 +1,65 @@
+export const createWasiBinding = (
+  localName: string,
+  wasiRegisterFunctions: string[],
+) => `/* tslint:disable */
```
forehalo: not related with this PR, but I think it's better to remove `tslint` ignore header?

cli/src/api/load-wasi-template.ts
```diff
@@ -0,0 +1,65 @@
...
+/* eslint-disable */
+/* prettier-ignore */
+
+/* auto-generated by NAPI-RS */\n
```
forehalo: wrong '\n' rendered

PR #1672, opened 2023-07-27
fix(napi): js promise error messages

stefan-gorules:
Adds more detailed error messages when promises fail in threadsafe function by trying to coerce them to a string. This is useful in situation when an error is thrown within async function, and enables capturing of errors in such cases.

For example (in https://github.com/gorules/zen):
```ts
const { ZenEngine } = require('./index');

const engine = new ZenEngine({
    loader: async (key) => {
        throw Error("Custom thrown error");
    }
});

(async () => {
    try {
        await engine.evaluate("unknown key", {});
    } catch (e) {
        console.log(e);
    }
})();
```

Fails without "Custom thrown error" until changes above are introduced. When introduced stack is visible with e.g.:
```
[Error: Loader error: Internal error on key unknown key: GenericFailure, Error: Custom thrown error]
```

### DIFF
--- a/crates/napi/src/error.rs
+++ b/crates/napi/src/error.rs
@@ -90,6 +90,18 @@ impl From<JsUnknown> for Error {
         "Create Error reference failed".to_owned(),
       );
     }
+
+    let maybe_error_message = value
+      .coerce_to_string()
+      .and_then(|a| a.into_utf8().and_then(|a| a.into_owned()));
+    if let Ok(error_message) = maybe_error_message {
+      return Self {
+        status: Status::GenericFailure,
+        reason: error_message,
+        maybe_raw: result,
+      };
+    }
+
     Self {
       status: Status::GenericFailure,
       reason: "".to_string(),


stefan-gorules:
Also @Brooooooklyn if there are cleaner ways of propagating error in above scenario, happy to have a look into it with some small direction. Ty

stefan-gorules:
Still facing the same issue @Brooooooklyn after updating to:
```toml
napi = { version = "2.13.3", features = ["serde-json", "error_anyhow", "tokio_rt"] }
```

The code in question is here: https://github.com/gorules/zen/blob/36154b0054b952f9d7c1837b63b1c41f904dcb4d/bindings/nodejs/src/loader.rs#L54-L57, and the JS code used to replicate is:

Add details in loader for debug
```rust
        let result = promise.await.map_err(|e| {
            println!("Error data: {:?}", e);

            return LoaderError::Internal {
                key: key.to_string(),
                source: e.into(),
            };
        })?;
```

e.g. test.js + `node test.js`
```js
const {ZenEngine} = require('./index');

const engine = new ZenEngine({
    loader: async (str) => {
        throw Error("Custom error data");
    }
});

(async () => {
    await engine.evaluate("hello", {});
})();
```

Issue #1689, opened 2023-08-08
Allow generation of non const enums


samvdst:
`@napi-rs/cli` always generates `const enum`. The build flag `--const-enum` defaults to true, which means there's no way to disable it. If you don't set the flag, it's true, and if you set it, it's also true.

I have seen the issue #1345 and also PRs #1527, #1647. None of them resolved the issue for me.

RobinMalfait:
Hey!

Just ran into this issue myself as well. Looking at the library they use ([clipanion](https://mael.dev/clipanion/docs/api/option#optionboolean)) it looks like setting `--const-enum false` is not a real thing. Instead passing `--no-const-enum` solves the issue for me. Hope that helps!

samvdst:
Hey @RobinMalfait, thanks for the tip. I basically created this PR because I couldn't find a way to negate these options in the documentation and help messages. Ideally, we should update the documentation to include these negated options. I haven't found the time to do that yet.

PR #1690, opened 2023-08-08
fix(cli): allow disabling of const enums (#1689)

samvdst:
This is a fix for #1689

### DIFF
--- a/cli/src/def/build.ts
+++ b/cli/src/def/build.ts
@@ -62,6 +62,11 @@ export abstract class BaseBuildCommand extends Command {
       'Path and filename of generated type def file. Relative to `--output_dir`',
   })
 
+  noConstEnum?: boolean = Option.Boolean('--no-const-enum', {
+    description:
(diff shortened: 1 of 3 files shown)


cli/src/def/build.ts
```diff
@@ -62,6 +62,11 @@ export abstract class BaseBuildCommand extends Command {
       'Path and filename of generated type def file. Relative to `--output_dir`',
   })
 
+  noConstEnum?: boolean = Option.Boolean('--no-const-enum', {
```
Brooooooklyn, line 65:
```suggestion
  constEnum?: boolean = Option.Boolean('--const-enum', {
```

and the `--no-const-enum` option will be auto generated

samvdst: 🤯 I didn't know it generated `--no-const-enum`. I googled for a very long time.
I tried `--const-enum=false`, `--const-enum=0`, `--const-enum false`. 

It would be great to document this in the CLI usage help. Is there a way to automatically generate the negated documentation as well, or should I add it manually?

PR #1812, opened 2023-11-14
chore(cli): simplify publish step in ci-template

segevfiner:
This uses the standard way to handle npm auth in GitHub Actions.

Another improvement is to look for a tag push instead of a commit that has some line that matches a the version regex thing:
* Changing the check to `if echo "${{ github.ref_name }}" | grep "^v[0-9]\+\.[0-9]\+\.[0-9]\+$";` or even `if [[ ${{ github.ref_name }} =~ ^v[0-9]\+\.[0-9]\+\.[0-9]\+$ ]];`
* Turning `tags-ignore: **` to `tags: v[0-9]*`.
* And we can then also `if` then entire publish job with `if: github.ref_type == 'tag'` to save some GitHub Actions runtime.

`then` can also be in the same line as the `if`, but that's a matter of taste.

### DIFF
--- a/cli/src/api/templates/ci-template.ts
+++ b/cli/src/api/templates/ci-template.ts
@@ -543,6 +543,7 @@ jobs:
         with:
           node-version: 18
           cache: 'yarn'
+          registry-url: 'https://registry.npmjs.org'
 
       - name: 'Install dependencies'
         run: yarn install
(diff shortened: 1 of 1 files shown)


Brooooooklyn (approved):
Thanks

PR #1909, opened 2024-01-18
fix(wasm-runtime): wasm compatible with core-js

fz6m:
Some of the tc39 [proposal collection methods](https://github.com/tc39/proposal-collection-methods) ( `esnext` )  return different values than expected in Node.js, causes `esnext.map.delete-all` / `esnext.set.delete-all` etc. not to work with `@napi-rs/wasm-runtime`.

Refer:

 - https://github.com/oven-sh/bun/pull/7101
 - https://github.com/oven-sh/bun/pull/4493#issuecomment-1806825114
 - https://github.com/nodejs/node/blob/32ce1834c03a38397db2cf42137b9376f110733e/lib/internal/per_context/primordials.js#L383
 - https://github.com/zloirock/core-js/pull/1302
 - https://github.com/tc39/proposal-collection-methods/issues/2


Brooooooklyn:
@fz6m hey, thanks for your contribution! Can you patch the file using the `yarn patch` command like https://github.com/napi-rs/napi-rs/blob/main/.yarn/patches/buffer-npm-6.0.3-cd90dfedfe.patch

fz6m:
I'm not sure yarn can patch Node.js internal code, since it from `node:util`

Brooooooklyn:
The code you patch seems like from https://github.com/napi-rs/napi-rs/blob/main/.yarn/patches/node-inspect-extracted-npm-3.0.0-f661b6c334.patch#L75, you can modify it directly to avoid an additional patch command

fz6m:
Thanks for your reply, I reworked it.

PR #1945, opened 2024-02-11
feature: Add ability to represent nullable TypeScript fields

### LINKED ISSUE #1944: feature request: Add ability to represent nullable TypeScript fields
## Problem
Let's say you want to handle some HTTP requests (or whatever), whose payload contains a nullable field:
```typescript
interface Payload {
    id: number
    name: string | null
}

function do_stuff(payload: Payload) {}
```

And you want to accept these kinds of payloads:

```typescript
{id: 1, name: null}
{id: 2, name: 'kitan'}
```

The equivalent implementation in Rust would be something like this:

```rust
#[napi(object)]
pub struct Payload {
    pub id: u32,
    pub name: Option<String>,
}

#[napi]
pub fn do_stuff(payload: Payload) {}
```

But napi exports the following type definition,

```typescript
export interface Payload {
    id: number
    name?: string
}

export function do_stuff(payload: Payload): void
```

so it produces an error when `null` is passed in the optional field:
```typescript
// Error: Failed to convert JavaScript value `Null` into rust type `String`
do_stuff({id: 1, name: null})
```

Given that such use cases are pretty common, I think napi should also provide a way to handle `null`s in JavaScript objects.

## Proposal

To handle such cases, I suggest that napi exports the following type definition from the Rust code above:

```typescript
export interface Payload {
    id: number
    name?: string | null
}

export function do_stuff(payload: Payload): void
```

So it works in both cases:

```typescript
do_stuff({id: 1, name: null})     // Payload {id: 1, name: None}
do_stuff({id: 2, name: 'kitan'})  // Payload {id: 2, name: Some("kitan")}
```

sup39:
- Closes #1944


sup39:
Closed in favor of #1971

PR #1949, opened 2024-02-14
Ensure cwd and outputDir are used for finding files

andymac4182:
Using the latest alpha I had an issue using a non standard `--output-dir` with the universalize command where the JS would find the files but the XCode tool couldn't. 

I have a patch running locally with this change that resolved the issue. https://github.com/wypes-rs/wypes/blob/main/.yarn/patches/%40napi-rs-cli-npm-3.0.0-alpha.38-93961b489d.patch


Brooooooklyn (approved):
Thanks!

Issue #1948, opened 2024-02-13
Webpack/esbuild/bundler support or examples.


everett1992:
Webpack, esbuild, or other bundlers are somtimes used outside a browser context, for example packaging for Lambda. 

I couldn't find any examples of using a napi-rs library with webpack or esbuild  and my initial attempts to use [webpack's node-loader](https://webpack.js.org/loaders/node-loader/) or [esbuild's copy loader](https://esbuild.github.io/content-types/#external-file) did not work due to the module resolution code generated by napi-rs

```typescript
localFileExisted = existsSync(
  join(__dirname, 'snappy.linux-x64-gnu.node')
)
try {
  if (localFileExisted) {
    nativeBinding = require('./snappy.linux-x64-gnu.node')
  } else {
    nativeBinding = require('@napi-rs/snappy-linux-x64-gnu')
  }
} catch (e) {
  loadError = e
}
```

Both bundlers correctly replaced this line with the file path from their bundle
```typescript
    nativeBinding = require('./snappy.linux-x64-gnu.node')
```

But they did not replace `localFileExisted` because it does not use require or import.
```typescript
localFileExisted = existsSync(
  join(__dirname, 'snappy.linux-x64-gnu.node')
)
```

In both webpack and esbuild the existsSync check would fail because the bundled code included a hash in the file path or relocated the file. I was able to get esbuild working with the following options
```typescript
import * as esbuild from 'esbuild'

await esbuild.build({
  // Copy .node files into outdir and update imports to the bundled path.
  loader: { '.node': 'copy' }
  // Do not add hash to bundled files, so napi-rs's localFileExisted check succeeds. 
  assetNames: '[name]',
})

```

everett1992:
I'm not sure why napi-rs generates the code it does, instead of maybe a try/catch, or a loop over candidate files. I can work on a PR if either of those sound good.

It would benefit me if the code could be tree shook and platform/arch could be overwritten  so I can create a bundle from a mac that will run on linux and have the big `switch (platform)` boil down to one  require for the target platform.

andymac4182:
Sounds good to me :) 

It would also be good if this code could be using both commonjs and esmodule. ESM support would help a lot with tree shaking.

Brooooooklyn:
@everett1992 I'll add some examples in the next week

Brooooooklyn:
@everett1992 see this example https://github.com/napi-rs/use-napi-rs-packages-with-bundlers/actions/runs/7991162667

[...]

PR #1957, opened 2024-02-17
refactor(cli): refactor js-binding to support easier bundling.

everett1992:
https://github.com/napi-rs/napi-rs/issues/1948

This PR addresses two issues I had with using the JavaScript generated by @napi-rs/cli with a bundler.
1. The existsSync check before require breaks if the bundler renames the .node file
2. platform, arch cannot be defined at build time.

With this change `esbuild` can bundle napi-rs libraries with just `--loader:.node=copy` and webpack can bundle with node-loader.
Before I was able to use esbuild by disabling asset hashing, but I wasn't able to get webpack working.

Additionally you can define process.arch and process.platform at build time and requires to unused platforms/arch's will be eliminated and not included in the bundle.


Brooooooklyn:
NAPI-RS had a similar implementation before; but it breaks another tools like `ncc` on Vercel: https://github.com/napi-rs/node-rs/issues/316

everett1992:
Do you have a link to that similar implementation?

I tested this PR with @vercel/ncc, esbuild, and webpack with both grouped and separated package binaries.
This implementation seems more likely to work across all bundlers because it doesn't use any `require.resolve` or file system checks. It's the same pattern you're currently using for separated package binaries, but I use it for both separated and grouped, and skip the file exists check. 

```typescript
 function requireNative() {
  if (process.platform === 'android') {
    if (process.arch === 'arm64') {
      try {
        return require('./rust-cloudauth-core.android-arm64.node')
      } catch (e) {
        loadErrors.push(e)
      }
      try {
        return require('@amzn/rust-cloudauth-core-android-arm64')
      } catch (e) {
        loadErrors.push(e)
      }

```

Brooooooklyn:
> Do you have a link to that similar implementation?

https://github.com/napi-rs/node-rs/blob/main/packages/helper/src/loader.ts

For loading logic like this:
```js
function requireNative() {
  if (process.platform === 'android') {
    if (process.arch === 'arm64') {
      try {
        return require('./rust-cloudauth-core.android-arm64.node')
      } catch (e) {
        loadErrors.push(e)
      }
      try {
        return require('@amzn/rust-cloudauth-core-android-arm64')
      } catch (e) {
        loadErrors.push(e)
      }
```

I'm worried that such logic is too dynamic, some packing tools that are not implemented finely may not be able to analyze the actual path/package required.

everett1992:
Which part are you worried about?

I could remove the requireNative function.
try require catch is what you're currently doing. I changed to an array of
errors, but could revert.
I think this is less dynamic because it removes a runtime fs exists check.

PR #1966, opened 2024-02-21
feat(napi): allow Reference as a class method param

lbarthon:
As of before this commit, there was a lock in the codegen preventing Reference from being used as a function argument outside of a Reference<Self>.

This changes it, allowing Reference of any class to be added as a class method argument anywhere. It has the same limitations as reference, as in it requires the class to have been created with a factory or constructor. This change implements FromNapiValue on Reference, which will unwrap the class and call the existing from_value_ptr method. It also updated typegen so that we only emit the reference type if we're in an impl block that doesn't match the Reference we're getting. This ensures that typegen works as expected with the previous behaviour.


Brooooooklyn (approved):

Brooooooklyn (commented):
Can you add a test for this scenario?

Brooooooklyn (approved):

PR #1968, opened 2024-02-22
fix(napi): JsObject.add_finalizer callback being allowed to reference non-static things

tomb-msft:
The `finalize_cb` passed to `add_finalizer` doesn't have a lifetime requirement, so it's allowed to reference things in scope. For example, this compiles fine, but it will segfault at runtime when the finalizer is called:

```rs
use napi::{JsObject, Result};

fn my_function(object: &mut JsObject) -> Result<()> {
    let my_string = String::from("Hello, world!");

    object.add_finalizer((), (), |_ctx| {
        println!("{}", my_string);
    });

    drop(my_string);
}
```

This commit just requires the callback has a static lifetime which prevents the example above compiling. However the other two arguments already require static lifetimes so I'm not sure why the callback didn't.

### DIFF
--- a/crates/napi/src/js_values/object.rs
+++ b/crates/napi/src/js_values/object.rs
@@ -37,7 +37,7 @@ impl JsObject {
   where
     T: 'static,
     Hint: 'static,
-    F: FnOnce(FinalizeContext<T, Hint>),
+    F: FnOnce(FinalizeContext<T, Hint>) + 'static,
   {
     let mut maybe_ref = ptr::null_mut();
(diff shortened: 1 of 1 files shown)


Brooooooklyn (approved):
Thanks!

Brooooooklyn:
### Merge activity

* **Feb 22, 5:26 AM EST**: [Graphite](https://app.graphite.dev/github/pr/napi-rs/napi-rs/1968) failed to merge this pull request due to **internal failures**. Try your merge again, or report a bug if you see this consistently.

Issue #1665, opened 2023-07-20
FATAL ERROR: threadsafe_function.rs:720


Aitthi:
Can I ignore this error?

```rs
#![deny(clippy::all)]

#[macro_use]
extern crate napi_derive;
use std::collections::HashMap;

use napi::{
  bindgen_prelude::*,
  threadsafe_function::ThreadsafeFunction
};
use axum::{
  routing::get,
  Router, response::Response, http::StatusCode, body::Body,
};


#[napi]
pub async fn serve(routes: HashMap<String, ThreadsafeFunction<String>>) -> Result<bool> {

    let mut app = Router::new();

    for (path, callback) in routes {
      app = app.route(path.as_str(), get(|| async move { 
        let mut res = Response::builder();
        let js_res = match callback.call_async::<serde_json::Value>(Ok("".to_string())).await {
            Ok(res) => res,
            Err(e) => {
                println!("Error: {}", e);
                serde_json::Value::Null
            }
        };
        let js_to_string = js_res.to_string();
        let mut js_data = js_to_string.as_str();
        if js_res.is_string() {
          js_data = js_res.as_str().unwrap_or("");
          res = res.header("Content-Type", "text/html");
        } else {
          res = res.header("Content-Type", "application/json");
        }
        res = res.status(StatusCode::OK);
        let body = res.body(Body::from(js_data.as_bytes().to_vec()));
        body.unwrap()
      }));
    }

    axum::Server::bind(&"0.0.0.0:3000".parse().unwrap())
        .serve(app.into_make_service())
        .await
        .unwrap();
  Ok(true)
}
```

```sh
FATAL ERROR: threadsafe_function.rs:720 Failed to convert return value in ThreadsafeFunction callback into Rust value: GenericFailure, Failed to send return value to tokio sender
 1: 0x10028b984 node::Abort() [/Users/Aitthi/.n/bin/node]
 2: 0x10028ba7c node::OOMErrorHandler(char const*, bool) [/Users/Aitthi/.n/bin/node]
 3: 0x10028b99c node::OnFatalError(char const*, char const*) [/Users/Aitthi/.n/bin/node]
 4: 0x10025f020 napi_open_callback_scope [/Users/Aitthi/.n/bin/node]
 5: 0x10cf36ae0 napi::threadsafe_function::call_js_cb::h09f46e41d2b76c8e [/Users/Aitthi/code/nlab/nlab.darwin-arm64.node]
[...]

PR #1970, opened 2024-02-22
fix(napi): no hard fail on ThreadsafeFunction::call_async

lbarthon:
For many reasons, in an app we can end up in a scenario where the thread / task that initially called the ThreadsafeFunction gets interrupted. This means that the receiver would get collected, should a ThreadsafeFunction be awaited on the said thread. This will create an error when the ThreadsafeFunction's callback will be called, and call a napi_fatal_error.

This change makes it so that if the send errors, this error is hidden and thus prevents any hard failure. If a ThreadsafeFunction is called in such a case, its output won't be used, but we'll still run all the logic to ensure it ran properly.

Fixes https://github.com/napi-rs/napi-rs/issues/1665 - cc @Brooooooklyn

### DIFF
--- a/crates/napi/src/threadsafe_function.rs
+++ b/crates/napi/src/threadsafe_function.rs
@@ -494,9 +494,9 @@ impl<T: 'static> ThreadsafeFunction<T, ErrorStrategy::CalleeHandled> {
                 callback: Box::new(move |d: Result<JsUnknown>| {
                   sender
                     .send(d.and_then(|d| D::from_napi_value(d.0.env, d.0.value)))
-                    .map_err(|_| {
-                      crate::Error::from_reason("Failed to send return value to tokio sender")
-                    })
+                    // The only reason for send to return Err is if the receiver isn't listening
+                    // Not hiding the error would result in a napi_fatal_error call, it's safe to ignore it instead.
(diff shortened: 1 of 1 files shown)


crates/napi/src/threadsafe_function.rs
```diff
@@ -494,9 +494,7 @@ impl<T: 'static> ThreadsafeFunction<T, ErrorStrategy::CalleeHandled> {
                 callback: Box::new(move |d: Result<JsUnknown>| {
                   sender
                     .send(d.and_then(|d| D::from_napi_value(d.0.env, d.0.value)))
-                    .map_err(|_| {
-                      crate::Error::from_reason("Failed to send return value to tokio sender")
-                    })
+                    .or(Ok(()))
```
Brooooooklyn, line 499: Maybe it's worth to add comments here to explain that it's okay to ignore the `send error`

lbarthon: On it

lbarthon: Done - mind tagging once merged? 🙏

lbarthon:
Thank you so much for merging @Brooooooklyn - do you know when you'll tag next?

PR #1991, opened 2024-03-07
feat(napi): support returning raw value for threadsafe function

h-a-n-a:
Support custom handling for returned value from a threadsafe function. 
This is helpful for which Rust projects want to handle errors.

For example, in Rspack, we will grab the `Error` thrown on JS side and convert it to better displayed diagnostics.
Ref: https://github.com/web-infra-dev/rspack/pull/5885/files#diff-fd19e247d09f6fb1b0b7431ecd354604951d29fc86e5c8e434acb56fb307fa60R50


h-a-n-a:
Closed in favor of https://github.com/napi-rs/napi-rs/pull/1997

PR #2010, opened 2024-03-22
fix(napi): no panic when caller stops listening

lbarthon:
This fix is similar to the one in https://github.com/napi-rs/napi-rs/commit/5b5f616d81ff2f0e154ff8b1e1edee1dfc087eb4.

In both the then_callback & catch_callback, expect was being called in case the send failed. This means that if we call a function that returns a promise and the received gets closed (the calling thread stopped at the wrong time), this will panic.

In such scenarios, it is fine not to panic. If the receiver doesn't care about the output, we should just let it be.


lbarthon:
fyi @Brooooooklyn

lbarthon:
Hey @Brooooooklyn - sorry for the pings - any idea when you'll be able to tag on top of that?

PR #2019, opened 2024-04-01
feat(target): add support for armv7-unknown-linux-muslebihf

sapphi-red:
This PR adds support for armv7 + musl using zig.


.github/workflows/test-release.yaml
```diff
@@ -401,7 +401,7 @@ jobs:
             ~/.cargo/registry
             ~/.cargo/git
             target
-          key: stable-linux-armv7-gnu-node@18-cargo-cache
+          key: stable-linux-armv7-gnueabihf-node@20-cargo-cache
```
sapphi-red, line 404: I didn't add a test for armv7-unknown-linux-muslebihf here because `@napi-rs/cross-toolchain` doesn't support it.

.github/workflows/zig.yaml
```diff
@@ -130,13 +133,23 @@ jobs:
...
+      - name: Test
+        uses: addnab/docker-run-action@v3
+        if: matrix.settings.target == 'armv7-unknown-linux-musleabihf'
+        with:
+          # https://github.com/nodejs/docker-node/issues/1798
+          image: node:18-alpine
+          options: --platform linux/${{ matrix.settings.docker-platform }} -v ${{ github.workspace }}:/build -w /build
+          run: |
+            set -e
+            yarn test:without-lerna
```
sapphi-red, line 152: lerna (nx) doesn't support armv7 so I had to run it directly with yarn. I used node 18 for this test because when using node 20 `yarn test` got stuck (https://github.com/nodejs/docker-node/issues/1798).

PR #2020, opened 2024-04-01
feat(target): add support for armv7-unknown-linux-muslebihf

sapphi-red:
This is a backport of #2019


.github/workflows/zig.yaml
```diff
@@ -61,8 +60,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v3
         with:
-          # Testing for compatibility with node v12.x
-          node-version: 12
+          node-version: 18
```
sapphi-red, line 63: yarn v4 only supports Node.js 18+. I bumped the node version here. Even with downgrading yarn, the test wasn't passing.

PR #2035, opened 2024-04-13
chore: increase wasi memory limits

Brooooooklyn:


Brooooooklyn:
* **#2035** <a href="https://app.graphite.dev/github/pr/napi-rs/napi-rs/2035?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @Brooooooklyn and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

cli/src/api/templates/load-wasi-template.ts
```diff
@@ -24,8 +24,10 @@ const __wasi = new __WASI({
 const __emnapiContext = __emnapiGetDefaultContext()
 
 const __sharedMemory = new WebAssembly.Memory({
-  initial: 1024,
-  maximum: 10240,
+  // 1Gb
+  initial: 16384,
```
toyobayashi, line 28: The `initial` size means `new WebAssembly.Memory()` will allocate 1GB size `SharedArrayBuffer` immediately, I personally don't think most of normal wasm module need so much initial memory, it just grow within `maximum` on demand. If there are multiple napi-rs modules are using in a same web page or a StackBlitz Node.js worker, is this too waste?

cli/src/api/templates/load-wasi-template.ts
```diff
@@ -95,8 +97,10 @@ const __wasi = new __nodeWASI({
...
```
toyobayashi, line 101: Same here

cli/src/api/templates/load-wasi-template.ts
Brooooooklyn: Good point. Let's make it configurableble

PR #2039, opened 2024-04-15
fix(napi): ensure CalleeHandled works as expected

lbarthon:
This was refactored in 4719caa64377f7d926c92a4c1051474ae79036c4. This codepath was swapped, as the fatal exception must be the normal codepath when the callee handled field is false. This fixes it by swapping this if statement, all the other calls have been checked and seem fine.

### DIFF
--- a/crates/napi/src/threadsafe_function.rs
+++ b/crates/napi/src/threadsafe_function.rs
@@ -658,7 +658,7 @@ unsafe extern "C" fn call_js_cb<
           )
         },
         Err(e) => {
-          if CalleeHandled {
+          if !CalleeHandled {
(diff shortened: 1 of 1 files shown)


Brooooooklyn (approved):
Oh, nice catch

lbarthon:
No problem, mind tagging on top of that? 😄 
@Brooooooklyn

Brooooooklyn:
Published: https://github.com/napi-rs/napi-rs/releases/tag/napi%402.16.3

Brooooooklyn:
Oh my bad 😥 I'll yank the `2.16.3` and publish a new alpha version

Issue #1480, opened 2023-02-08
Is it possible to build node binaries for nwjs on windows?


MikalDev:
I am new to rust / napi-rs. I am trying to see if it is possible to build new binaries to work with nwjs, instead of node/electron (I am using a build service which uses nwjs.)

I have tried nwjs with the npm node files on MacOS and it seems to work fine, with various versions of nwjs.

However on Windows, I can only get electron versions of my app to work.

When I try using the node files with nwjs on Windows, the app crashes when loading the node binary.

I am able to build the node binaries on Windows, but I am not sure if I need to do something specific to get it compiled for nwjs for windowsI have seen comments about the possibility of different native node modules requiring different compiles on node vs electron vs nwjs as they may have different ABIs.

Brooooooklyn:
NAPI-RS should work with nwjs, could you provide a minimal reproducible repo so I can debug it?

MikalDev:
I will look at building a min repro repo. (Where I see this happen now is with https://github.com/ceifa/steamworks.js which has a major dependency of https://github.com/Noxime/steamworks-rs, I will look at doing something more manageable. )

One quick question - when building for nwjs, are there any specific napi-rs parameters or build config that needs to be done?)

GMMan:
I've looked into this, and the problem is the loading code assumes the host process module is where the `napi` functions are exported (by calling `libloading::os::windows::Library::this()`), rather than actually finding the Node runtime. NW.js loads Node as a library, so no `napi` functions are available on `nw.exe` itself. This causes `panic_load()` to be called, which probably isn't caught anywhere and causes the renderer process to crash.

MikalDev:
Really interesting. Do you have repo that you can share, so others can debug?

GMMan:
I was using Steamworks.js too. Here's a repro repo: https://github.com/GMMan/nw-napi-rs-crash

MikalDev:
@Brooooooklyn - is the above repo enough for you to take a look at debugging? Need any other info?

Brooooooklyn:
@MikalDev yes, sure. I'll find some time to debug it

MikalDev:
@Brooooooklyn - bumping this again to see if it will possible to debug.

MikalDev:
Lots of speculation here, I am new to Rust.

Looking at this more with @GMMan 's comment, I see the change where it went from using 

[...]

PR #2042, opened 2024-04-15
Enable windows nw.js to work. Change to look for external node.dll as a fallback for Windows platform

MikalDev:
This is to allow for node files to also work with Windows nw.js. not just electron and node.

Using the patch from this issue: #1480.

The patch checks if a node library function is available in the executable, if not, it loads the external library version of node via node.dll (which is already loaded.)

### DIFF
--- a/crates/sys/src/functions.rs
+++ b/crates/sys/src/functions.rs
@@ -780,10 +780,37 @@ pub use napi8::*;
 #[cfg(feature = "napi9")]
 pub use napi9::*;
 
+#[cfg(windows)]
+fn test_library( lib_result : Result<libloading::os::windows::Library, libloading::Error> ) -> Result<libloading::Library, libloading::Error> {
+  unsafe {
+    match lib_result {
+      Ok(lib) => {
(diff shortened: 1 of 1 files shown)


crates/sys/src/functions.rs
```diff
@@ -780,10 +780,37 @@ pub use napi8::*;
...
+  return test_library(libloading::os::windows::Library::this())
+    .or(test_library(libloading::os::windows::Library::open_already_loaded("node")));
+}
+
 #[cfg(any(windows, feature = "dyn-symbols"))]
 pub(super) unsafe fn load_all() -> Result<libloading::Library, libloading::Error> {
   #[cfg(windows)]
-  let host = libloading::os::windows::Library::this()?.into();
+  let host_result = find_node_library();
+  let host : libloading::Library;
```
lbarthon: `let host = find_node_library()?;` - instead of those 6 lines?

MikalDev: Agree, updated commit.

MikalDev:
Ah, should not have used the update branch option on github, it caused an unsigned commit, will create a new PR.

PR #2052, opened 2024-04-18
Add support for nwjs adding fallback for node library search

MikalDev:
This is to allow for node files to also work with Windows nw.js. not just electron and node.

Using the patch from this issue: https://github.com/napi-rs/napi-rs/issues/1480.

The patch checks if a node library function is available in the executable, if not, it loads the external library version of node via node.dll (which is already loaded.)


MikalDev:
Sigh, fail again with signing verification. Will try again.

PR #2075, opened 2024-04-30
fix: support windows gnu

ahaoboy:


ahaoboy:
> shouldnt you detect if the node you are using is gnu or msvc?

Since the rust toolchain version is modified by `rustup override set nightly-x86_64-pc-windows-gnu` or ` cargo build --target=nightly-x86_64-pc-windows-gnu`, I did not find a way to get this information in node. Any ideas?

Kreijstal:
> Since the rust toolchain version is modified by `rustup override set nightly-x86_64-pc-windows-gnu` or ` cargo build --target=nightly-x86_64-pc-windows-gnu`, I did not find a way to get this information in node. Any ideas?

https://stackoverflow.com/questions/78610813/how-to-tell-if-node-in-windows-is-the-msvc-version-or-the-mingws-version/78610827#78610827

ahaoboy:
> > Since the rust toolchain version is modified by `rustup override set nightly-x86_64-pc-windows-gnu` or ` cargo build --target=nightly-x86_64-pc-windows-gnu`, I did not find a way to get this information in node. Any ideas?
> 
> https://stackoverflow.com/questions/78610813/how-to-tell-if-node-in-windows-is-the-msvc-version-or-the-mingws-version/78610827#78610827


This method can only get the system version, not the target version of the build product. In the case of os=Windows_NT, by switching the rust toolchain, I can build products in both msvc and gnu formats.
```js
const os = require('os');
console.log(os.type());
```

Kreijstal:
> > > Since the rust toolchain version is modified by `rustup override set nightly-x86_64-pc-windows-gnu` or ` cargo build --target=nightly-x86_64-pc-windows-gnu`, I did not find a way to get this information in node. Any ideas?
> > 
> > 
> > https://stackoverflow.com/questions/78610813/how-to-tell-if-node-in-windows-is-the-msvc-version-or-the-mingws-version/78610827#78610827
> 
> This method can only get the system version, not the target version of the build product. In the case of os=Windows_NT, by switching the rust toolchain, I can build products in both msvc and gnu formats.
> 
> ```js
> const os = require('os');
> console.log(os.type());
> ```

you mean, you want to target GNU's node from MSVC node? 
Hmm, interesting, I don't know how compatible they are.

ahaoboy:
> you mean, you want to target GNU's node from MSVC node?


I don't know the specific difference between msvc and gnu. My default environment is gnu, so the build product format is xxx.win32-x64-gnu, which causes the loading script to not find the product. I try to change it to xxx.win32-x64-msvc and it can run.
Maybe it's because the program is simple. If it may cause other problems, I'll close this PR.

ahaoboy:
> I just assumed rust gnu uses nodejs gnu, never thought about rust gnu and node msvc for example, hmm 🤔


In this simple example, you only need to rename the package-template.win32-x64-gnu.node file to package-template.win32-x64-msvc.node to pass the test. Of course, this may also be because this example is too simple.
https://github.com/ahaoboy/napi-gnu-test/actions/runs/9484626299/job/26134544427

Issue #1541, opened 2023-03-24
@napi-rs/cli@2.15.2 generate duplicate definitions in .d.ts


xiaopengli89:

Brooooooklyn:
Try to upgrade `napi-derive`

xiaopengli89:
Thanks!

xiaopengli89:
The definition still be repeated.

napi-cli: 2.15.2
napi: 2.12.1
napi-derive: 2.12.2
napi-build: 2.0.1

miniben-90:
Hi,
I have same issue with duplication of .d.ts and .js code only in windows 11 (the code is duplicated 3 times for .d.ts and 6 times for index.js)
* rustc: 1.68.2 (9eb3afe9e 2023-03-27)
* cargo: 1.68.2 (6feb7c9cf 2023-03-26)
* NodeJS: v18.12.1
* @napi-rs/cli: 2.15.2
* napi: 2.12.0
* napi-derive: 2.12.2

Brooooooklyn:
@miniben-90 Because of the limitations of the Rust macro, your case `napi-derive` cannot currently handle... You may need to add `#[napi(skip_typescript)]` under 2 of these platforms

xiaopengli89:
I don't have conditional compilation code, it works fine with 2.14.8.

Brooooooklyn:
@xiaopengli89 great!

miniben-90:
> @miniben-90 Because of the limitations of the Rust macro, your case `napi-derive` cannot currently handle... You may need to add `#[napi(skip_typescript)]` under 2 of these platforms

Thanks for your answer, but nothing change after adding skip_typescript... I switch back to 2.14.8 for `@napi-rs/cli`.

Brooooooklyn:
@xiaopengli89 do you have a minimal reproducible repo?

forehalo:
> 



> I removed the rust binding method, but it still generates the previous code, and it looks like `@napi-rs/cli@2.15.2` is caching the previous code. This problem does not occur when creating a new project.

We do cache the previous typedef because users may delete `*.d.ts` generated previously but leave rust code untouched, in which case, cargo will not compile the rust code due to strong cache and we finally get an empty `*.d.ts`.

After removing rust binding method, you should compile your crate with `cargo build` directly instead of using `napi build` because it's no longer node binding lib and all the pre-build and post-build stuffs napi cli do are useless.

art049:
[...]

PR #2088, opened 2024-05-06
remove duplicate typedefs in `index.d.ts` from macro re-evaluation

OmarTawfik:
Type definitions are generated during `cargo build`, and written to a temporary typedefs file. Usually, they are only evaluated once, but they can also be evaluated more times, when modules are loaded through [conditional compilation](https://github.com/napi-rs/napi-rs/issues/1541), by using `#[path]` attributes, or various other code patterns required to get around [existing cargo limitations](https://github.com/rust-lang/cargo/issues/8628).

This causes the temp file to have duplicate entries, which results in an invalid `index.d.ts` generated, with twice the number of types, causing `tsc` to fail with many `TS2300` errors. The issue happens with the latest released `@napi-rs/cli` version `2.18.2`.

This localized fix here deduplicates any entries found in the file, and has no effect when there are none.

### DIFF
--- a/cli/src/build.ts
+++ b/cli/src/build.ts
@@ -750,7 +750,9 @@ async function processIntermediateTypeFile(
     return idents
   }
 
-  const allDefs = lines.map((line) => JSON.parse(line) as TypeDef)
+  // Deduplicate lines, to account for macros being possibly evaluated more than once:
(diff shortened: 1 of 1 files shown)


OmarTawfik:
cc @ Brooooooklyn

PR #2114, opened 2024-05-16
feat(napi): Function build_thfn support calledhandled mode

southorange0929:


Brooooooklyn:
You should make the `typegen` correct for the `calledhandled = true` tsfn

southorange0929:
I think we don't need to add some logic for `ThreadsafeFunctionBuilder` with `calleehandled = true`.  
It just is used to generate `ThreadsafeFunction`  which is support `typegen`, right?

Brooooooklyn (approved):

southorange0929:
Resolve conflict

PR #2115, opened 2024-05-16
chore(napi): allow value attribute on enum variants

lbarthon:
This allows us to customize the enum value for string_enum. So far, only the case attribute was allowing us to customize the value of those variants, this is a next step in allowing the value to be different from the actual name.

This has a new test, and snapshots have been updated accordingly.

@Brooooooklyn - could we include that in napi v2 and tag after it?


Brooooooklyn (approved):

Brooooooklyn:
> could we include that in napi v2 and tag after it?

sure

PR #2117, opened 2024-05-17
fix(typegen): add missing `declare` keyword

### LINKED ISSUE #2106: generated d.ts definitions should use `export declare class` instead of `export class`
**current behavior**: NAPI currently generates class declarations in the `d.ts` file that look like this:

```
export class Foo {
  foo1(): void
  get foo2(): void
}
```

This is not valid in a `class` definition.

**expected behavior**: a `class` declaration:

```
export declare class Foo {
  foo1(): void
  get foo2(): void
}
```

Not totally sure where a PR to fix this would go, I glanced at https://github.com/napi-rs/napi-rs/blob/b2239fd880fa40fa98d206d8f31aec1bb8a0ce12/crates/backend/src/typegen/struct.rs but didn't see an obvious answer

magic-akari:
- Close #2106


himself65:
Could you release a new patch version for @napi-rs/cli ? This will fix the function type decoration.

/cc @Brooooooklyn

Issue #2133, opened 2024-06-05
Difference in Strings with null chars in Behavior in compat mode and  normal mode.


sheki:
# Summary

Compat mode truncates strings passed to it when it encounters the first NULL charecter.

Likely the issue is in

```rust
impl JsStringUtf8 {
  pub fn as_str(&self) -> Result<&str> {
    unsafe { CStr::from_ptr(self.buf.as_ptr()) }
      .to_str()
      .map_err(|e| Error::new(Status::InvalidArg, format!("{}", e)))
  }
```

### Js Code
```javascript
const { hello, greet } = require('./index')

const str = ['a', '\0', 'b', '\0', 'c'].join('')

console.log(str === 'abc')
console.log(str.length)
console.log(str)
console.log('Passing in string with nulls, expect to print abc')
console.log('#[js_function] : ', hello(str))
console.log('#[js_function] length: ', hello(str).length)
console.log('#[napi] : ', greet(str))
console.log('#[napi] length ', greet(str).length)
```
### Rust code
```rust
#[napi]
fn greet(name: String) -> String {
	format!("{}", name)
}
#[js_function(1)]
fn hello(ctx: CallContext) -> Result<JsString> {
  let argument_one = ctx.get::<JsString>(0)?.into_utf8()?;
  ctx.env.create_string_from_std(format!("{}", argument_one.as_str()?)) ## The loss happens in the as_str implementation 
}

#[module_exports]
fn init(mut exports: JsObject) -> Result<() > {
  exports.create_named_method("hello", hello)?;
  Ok(())
}
```
### Output
```
false
5
abc
Passing in string with nulls, expect to print abc
#[js_function] :  a
#[js_function] length:  1
#[napi] :  abc
#[napi] length  5
```

Issue #2110, opened 2024-05-13
Unexpect `Nul` Error of CString while creating javascript object


hyf0:
Related:
- https://github.com/rolldown/rolldown/issues/1115

Rollup use '\0' as the virtual module convention, which some modules might have names like "\0some-virtual-modules". Rolldown follows the behavior, but there's a error occrured in such code: https://github.com/rolldown/rolldown/blob/49d4b5c03ab699d71125782c746d15b9b276821c/crates/rolldown/src/utils/chunk/render_chunk.rs#L76-L78

I guess this is due to NAPI-RS use `CString` on places doesn't require to be valid `CString`.

hyf0:
Might be related

https://github.com/napi-rs/napi-rs/blob/aeb0b4766d9554519eb4002e02910ebb2b4a0b36/crates/napi/src/js_values/mod.rs#L315

PR #2138, opened 2024-06-12
fix(napi): JsString should respect \0 character to align with String

southorange0929:
Resolve: https://github.com/napi-rs/napi-rs/issues/2133


southorange0929:
I tried to fix `impl FromNapiValue for &str` together, but some other problems arose. It seems to conflict with the logic of the https://github.com/napi-rs/napi-rs/issues/2110.

Brooooooklyn (approved):

Issue #2152, opened 2024-06-23
RFC: "async_local" feature that uses libuv to drive Rust futures on the local JavaScript thread


alshdavid:
HI all 👋  

This issue tracks the discussion for driving Rust futures on the local JavaScript thread (with a working implementation).

## Why?

Currently, trying to write asynchronous Nodejs functionality with napi necessitates dancing around with `ThreadsafeFunction`, `AsyncTask`, `Deferred` - or a combination of these. 

The core async APIs Nodejs-napi gives us are not very ergonomic and introduce all sorts of challenges when navigating `!Send` types. Thread safe function can easily result in deadlocks if the developer is not careful.

Napi-rs offers a solution to these poor ergonomics by combining Tokio on a different thread with macros, giving consumers an ergonomic API to work with that minimizes the annoyance of thread-management - however it is still fraught with challenges.

The goal of this local-thread-async feature is to give consumers the best of both worlds.  It would allow for async code to be written in Rust which can accept `!Send` types (like `Env`, `JsValue` types) where timers and channels can be used without blocking the JS thread.

Of course, it's still wise to ensure heavy computation happens in a dedicated thread. Having futures executed on the main thread allows for a simplified control flow where results can be sent back to the JavaScript thread via channels, replacing most use cases for thread safe functions.

## Implementation

I have a working implementation of this in a [fork of Neon](https://github.com/alshdavid-forks/neon/tree/libuv-powered-futures)

- [x] Local futures executor driven by libuv
- [x] `env.spawn_local()`
- [ ] `Ref` types for async closures 
- [ ] Macro for `#[neon(async_local)] async fn`
- [ ] Types generation
- [ ] Tests

## Examples

The examples below demonstrate the syntax achievable with this implementation _without_ blocking the main JS thread 

### Async Function Returning Promise

```rust
// Typedef:
//   function(env: Env, callback: Function): Promise<undefined>
#[napi(async_local)]
async fn foo(callback: JsFunction) -> Result<JsUndefined> {
  task::sleep(Duration::from_secs(1)).await;

  let arg = env.create_string("Hello World")?;
  callback.call(None, &[arg]);
  
  env.get_undefined()
}
```

### Async Closure Within Normal Function

```rust
// Typedef: 
//   function(env: Env, callback: Function): undefined
#[napi(async_local)]
async fn foo(callback: JsFunction) -> Result<JsUndefined> {
  // This is essentially setTimeout()
[...]

PR #2158, opened 2024-06-25
Non-blocking Rust futures

alshdavid:
This PR introduces a custom Rust futures runtime that runs on the main/local JavaScript thread without blocking the JavaScript event loop. 

This is gated behind the `async_local` feature and enables the use of async channels, timers and other similar patterns that simplify the development experience.

This is done by combining a custom futures executor with a thread-safe napi function.

The runtime supports utilities from any package that uses `std:future` - a recommendation is [async_std](https://github.com/async-rs/async-std).

```rust
#[napi]
pub fn run_callback() -> napi::Result<JsUndefined> {
  // Runs asynchronously on the main thread
  // without blocking the event loop
  env.spawn_local(|env| async move {
    task::sleep(Duration::from_secs(3)).await;
    let message = env.create_string("Hello World!")?;
    console_log(&env, &[message]);
  });

  env.get_undefined()
}
```

[More info](https://github.com/napi-rs/napi-rs/issues/2152)

## Future plans
- Add `env.spawn_local_weak` for unref'ed futures that shouldn't hold the process open
- Need a reference counted handle for JavaScript values that has a static lifetime
- Support for `async fn` with `#[napi(async_local)]` macro

### DIFF
--- a/crates/napi/Cargo.toml
+++ b/crates/napi/Cargo.toml
@@ -18,6 +18,7 @@ independent = true
 
 [features]
 async = ["tokio_rt"]
+async_local = ["dep:futures"]
 compat-mode = []
 default = ["napi3", "compat-mode"]                                               # for most Node.js users
 experimental = ["napi-sys/experimental"]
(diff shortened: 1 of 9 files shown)


alshdavid:
I have released this as a third party library: https://github.com/alshdavid/napi_async_local

PR #2188, opened 2024-07-15
fix: napi getter macros

SyMind:
image


Brooooooklyn:
@SyMind run `cargo fmt`

PR #2193, opened 2024-07-19
fix(cli): wasi load in woker_threads

Brooooooklyn:
Prevent the process.on('beforeExit') hook being called while NAPI-RS
wasi pacakge is loaded in worker_threads

### DIFF
--- a/cli/src/api/templates/load-wasi-template.ts
+++ b/cli/src/api/templates/load-wasi-template.ts
@@ -97,7 +97,7 @@ export const createWasiBinding = (
 const __nodeFs = require('node:fs')
 const __nodePath = require('node:path')
 const { WASI: __nodeWASI } = require('node:wasi')
-const { Worker } = require('node:worker_threads')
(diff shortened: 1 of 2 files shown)


Brooooooklyn:
* **#2193** <a href="https://app.graphite.dev/github/pr/napi-rs/napi-rs/2193?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @Brooooooklyn and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

toyobayashi:
If user's intention is to use the napi-rs wasi module only in their own workers (not esm loader), this will cause the worker to never exit and the registered napi cleanup hooks will not be executed.

I personally recommend use increaseWaitingRequestCounter only in esm loader worker, instead of any worker.

PR #2209, opened 2024-08-05
chore: implement serde::Serialize for Either

lbarthon:
This adds an implementation of serde::Serialize for all the Either types. Long term, we might want to use the either crate, but for now this makes this implementation usable with serde.

If this sgty @Brooooooklyn - can it be added to napi2? :)


Brooooooklyn (approved):

lbarthon:
Hey @Brooooooklyn - thanks for merging! Can this be added to napi2?

Issue #2245, opened 2024-09-06
Node exceptions using `Uint8Array` in `async fn`


wemeetagain:
I ran into what seems like a double-free when using `Uint8Array` in an `async fn`.
Unfortunately, I don't have a small reproducible test, but the scenario looked like the following (simplified to highlight the problem):
```rust
#[napi]
pub struct Stream {
  send: SendStream
}

#[napi]
impl Stream {
  #[napi]
  pub async unsafe fn write(&mut self, data: Uint8Array) -> Result<()> {
    self.send.write_all(&data).await.map_err(to_err)
  }
}
```

Later used like so:
```ts
async function sink (stream: Stream, source: AsyncIterable<Uint8Array>) {
  for await (const chunk of source) {
    await stream.write(chunk)
  }
}
```
In unit tests, everything works as it should. However, when integrated in a live application, it results in segfaults and other node stack traces:
eg:
```
#
# Fatal error in , line 0
# Check failed: backing_store->is_wasm_memory().
#FailureMessage Object: 0x7f2323e39950
----- Native stack trace -----
 1: 0xffab01  [node]
 2: 0x28b3dfb V8_Fatal(char const*, ...) [node]
 3: 0x15be061  [node]
 4: 0x15be085 v8::internal::BackingStore::~BackingStore() [node]
 5: 0x121f42b std::_Sp_counted_deleter<v8::internal::BackingStore*, std::default_delete<v8::internal::BackingStore>, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() [node]
 6: 0x13d5a82 v8::internal::ArrayBufferSweeper::SweepingJob::SweepListFull(v8::internal::ArrayBufferList*) [node]
 7: 0x13d5dc9 v8::internal::ArrayBufferSweeper::SweepingJob::Sweep() [node]
 8: 0x13d5fcc v8::internal::ArrayBufferSweeper::DoSweep(v8::internal::ArrayBufferSweeper::SweepingType, v8::internal::ThreadKind, unsigned long) [node]
 9: 0x13d66d9 v8::internal::ArrayBufferSweeper::RequestSweep(v8::internal::ArrayBufferSweeper::SweepingType, v8::internal::ArrayBufferSweeper::TreatAllYoungAsPromoted) [node]
10: 0x148b592 v8::internal::MarkCompactCollector::Finish() [node]
11: 0x14691ae v8::internal::Heap::MarkCompact() [node]
[...]

PR #2083, opened 2024-05-03
Fix panic in JsArrayBufferValue as_ref/as_mut with Rust 1.78

tronical:
In Rust 1.78, when debug assertions are enabled, slice::from_raw_parts panics if the provided data is a null pointer. This is possible through JsArrayBufferValue::new() as well as through JsArrayBuffer::into_value, when napi_get_arraybuffer_info returns a null pointer due to a zero length buffer.

PR #2258, opened 2024-09-17
fix(napi): nullptr handling in `TypedArray`s

Xanewok:
The #2083 equivalent but for TypedArrays. I looked into the value handling as well and it seem that we also manifest the raw pointer into value in the finalizer, so this probably also fixes https://github.com/napi-rs/napi-rs/issues/2245, as that issue relates to a panic in a destructor for the backing store but the data is null in our case.

I can verify, based on the `napi@2.16.6` that this fixes the regression test case added in the branch.

We could probably do something similar to `Buffer` i.e. handle a `Option<NonNull<_>>` rather than a raw pointer but for now this will suffice; any refactor can be done as a follow-up to the fix here.

Would it be possible to release the fix for the v2?


Brooooooklyn (approved):

Xanewok:
Thanks a lot!

Xanewok:
@Brooooooklyn I'm seeing that this was tagged as being released in 2.16.11 but looking at https://crates.io/crates/napi/ it seems that only 2.16.10 was released so far.

Would it be possible to release this to crates.io, as well, please? :pray:

PR #2284, opened 2024-09-30
feat(cli): support ESM binding

### LINKED ISSUE #2244: Support native ESM codegen
In ESM projects, where napi bindings are just one part of the repo, mixing ESM and CJS files can be a huge pain.
It would be nice for napi-rs to generate an index.js that is valid ESM.

At a bare minimum, this can be done by adding the following lines to the top of the generated file:
```js
import { createRequire } from 'node:module';
const require = createRequire(import.meta.url);

const __dirname = new URL('.', import.meta.url).pathname;
```
and updating the exports.

wemeetagain:
- Split `createCjsBinding` into reusable code chunks `createCommonBinding` and `bindingHeader`
- Add `createEsmBinding`
- Add a `--esm` option to build command to select esm vs cjs binding
- [x] Tested the ESM build locally
- resolves #2244


cli/src/def/build.ts
```diff
@@ -65,6 +65,11 @@ export abstract class BaseBuildCommand extends Command {
       'Whether to disable the generation JS binding file. Only works with `--platform` flag.',
   })
 
+  esm?: boolean = Option.Boolean('--esm', {
+    description:
+      'Whether to emit an ESM JS binding file. Only works with `--platform` flag.',
```
Brooooooklyn:
```suggestion
      'Whether to emit an ESM JS binding file instead of CJS format. Only works with `--platform` flag.',
```

Brooooooklyn (commented):
How about keep the cjs format (for cjs projects), and create the esm binding from cjs file:

```js
import { createRequire } from 'node:module';
const require = createRequire(import.meta.url);
const __dirname = new URL('.', import.meta.url).pathname;

export const { Foo, Bar, Baz } = require('./index.js') 
```

wemeetagain:
> and create the esm binding from cjs file

That could work. Although, the DX is annoying in that it would be another file to configure. Eg: `--js bindingFile.cjs --esm otherBindingFile.mjs`
I think its better to simply emit the proper kind of format straightaway.

PR #2285, opened 2024-09-30
feat!: support node 22

wemeetagain:
- Replace "import assertions" with "import attributes" (see https://nodejs.org/dist/v18.20.4/docs/api/esm.html#import-attributes)
  - Replacing import assertions breaks support for now-unsupported node 16
- Replace promisified `node:fs` functions with `node:fs/promises` functions
- Update supported engines to drop support for node 16 (which is not supported anymore see https://nodejs.org/en/about/previous-releases#release-schedule)

BREAKING CHANGE:
- Drop support for node 16

### DIFF
--- a/cli/esbuild.mjs
+++ b/cli/esbuild.mjs
@@ -1,7 +1,7 @@
 import * as esbuild from 'esbuild'
 import { pull } from 'lodash-es'
 
-import packageJson from './package.json' assert { type: 'json' }
+import packageJson from './package.json' with { type: 'json' }
 
 await esbuild.build({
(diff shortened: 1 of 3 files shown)


triples/package.json
```diff
@@ -48,7 +48,7 @@
   },
   "devDependencies": {
     "@napi-rs/cli": "workspace:*",
-    "@types/node": "^20.11.13",
+    "@types/node": "^22.7.4",
```
reneleonhardt, line 51: Can the dependencies be updated?

```suggestion
    "@types/node": "^22.17.0",
```

PR #2296, opened 2024-10-08
feat(napi): support arraybuffer for de

richerfu:
For Buffer and ArrayBuffer, we  need to handle it explicitly instead of using `FromNapiValue`.

Using `FromNapiValue` to convert value will get some error: 
image

### DIFF
--- a/crates/napi/src/js_values/de.rs
+++ b/crates/napi/src/js_values/de.rs
@@ -3,11 +3,13 @@ use std::convert::TryInto;
 use serde::de::Visitor;
 use serde::de::{DeserializeSeed, EnumAccess, MapAccess, SeqAccess, Unexpected, VariantAccess};
 
-use crate::bindgen_runtime::FromNapiValue;
+use crate::bindgen_runtime::{Buffer, FromNapiValue};
 #[cfg(feature = "napi6")]
 use crate::JsBigInt;
 use crate::{type_of, NapiValue, Value, ValueType};
 use crate::{Error, JsBoolean, JsNumber, JsObject, JsString, JsUnknown, Result, Status};
+
+use super::JsArrayBuffer;
 
 pub struct De<'env>(pub(crate) &'env Value);
 impl<'env> De<'env> {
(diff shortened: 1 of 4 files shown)


Brooooooklyn:
wasm test failed, seems like the emnapi `napi_is_arraybuffer` return false when value is `new ArrayBuffer(10)` /cc @toyobayashi

toyobayashi:
> seems like the emnapi napi_is_arraybuffer return false when value is new ArrayBuffer(10)

After I comment out this two line test passed. Seems like the `Uint8Array` is converted in [`FromNapiValue for Array`](https://github.com/richerfu/napi-rs/blob/b67b542afb15d8b9d8e3b6641e034e9c5589bfc9/crates/napi/src/bindgen_runtime/js_values/array.rs#L124)?

```ts
test('serde-buffer-bytes', (t) => {
  // t.is(testSerdeBufferBytes({ code: new Uint8Array([1, 2, 3]) }), 3n)
  // t.is(testSerdeBufferBytes({ code: new Uint8Array(0) }), 0n)

  t.is(testSerdeBufferBytes({ code: Buffer.from([1, 2, 3]) }), 3n)
  t.is(testSerdeBufferBytes({ code: Buffer.alloc(0) }), 0n)
  t.is(testSerdeBufferBytes({ code: new ArrayBuffer(10) }), 10n)
})
```

crates/napi/src/js_values/de.rs
```diff
@@ -83,14 +93,47 @@ impl<'x, 'de, 'env> serde::de::Deserializer<'x> for &'de mut De<'env> {
...
 
   fn deserialize_byte_buf<V>(self, visitor: V) -> Result<V::Value>
   where
     V: Visitor<'x>,
   {
-    visitor.visit_bytes(unsafe { FromNapiValue::from_napi_value(self.0.env, self.0.value)? })
+    match type_of!(self.0.env, self.0.value)? {
+      ValueType::Object => {
+        let js_object = unsafe { JsObject::from_raw_unchecked(self.0.env, self.0.value) };
+        if js_object.is_buffer()? {
```
toyobayashi, line 122: I see. Since native Node-API `napi_is_buffer(Uint8Array)` returns `true` but emnapi returns `false`, so the path here is different.

PR #2297, opened 2024-10-08
`createRequire` for SEA support

jackyzha0:
I'm currently using NAPI-RS with esbuild + Node.js's Single Executable Addons (SEAs), however the JS Bindings that NAPI-RS produces use raw `require` calls for local files.

With esbuild, this is normally fine as you can use the `copy` loader to just move the `.node` addon to the right place but SEAs dont allow requiring from the fs: https://nodejs.org/docs/latest-v20.x/api/single-executable-applications.html#requireid-in-the-injected-main-script-is-not-file-based

They recommend using `createRequire` instead

This change should allow native addons created by NAPI-RS to be used in SEAs


Brooooooklyn (dismissed):

wemeetagain:
- please consider https://github.com/napi-rs/napi-rs/pull/2284 as an additional alternative resolution

cli/src/api/templates/js-binding.ts
```diff
@@ -22,6 +22,8 @@ export function createCjsBinding(
 /* auto-generated by NAPI-RS */
 
 const { readFileSync } = require('fs')
+const { createRequire } = require('node:module')
+require = createRequire(__filename)
```
jackyzha0, line 27:
```suggestion
require = createRequire(__filename)
```

delete some whitespace so i can get github to sign a commit 🥴

cli/src/api/templates/js-binding.ts
```diff
@@ -22,6 +22,8 @@ export function createCjsBinding(
...
 
 let nativeBinding = null
```
jackyzha0:
```suggestion
require = createRequire(__filename)
let nativeBinding = null
```

delete some whitespace so i can get github to sign a commit 🥴

PR #2299, opened 2024-10-11
feat(napi): allow asynchronous to use custom status

richerfu:


richerfu:
Hi @Brooooooklyn  should we continue this work?  Or other suggestions? Thank you!

Issue #2246, opened 2024-09-09
Unable to make the typegen output custom TypeName implementations


JonoPrest:
For instance I want to be able to express an enum as a union of string literals. I've implemented TypeName and From/ToNapiValue.

The index.d.ts filed does not get the type declaration for "Test" in this case.

```rs
pub enum Test {
    A,
    B,
}

impl TypeName for Test {
    fn type_name() -> &'static str {
        "\"A\" | \"B\""
    }

    fn value_type() -> napi::ValueType {
        napi::ValueType::String
    }
}

#[napi]
pub fn test(val: Test) -> () {
    let _ = val;
}
```

Is there a way to make the build generate the types as defined in the trait TypeName trait?

CPunisher:
Can https://napi.rs/docs/concepts/enum#string-enum meet your requirements?

CPunisher:
You can use `npx napi build --platform --release --no-const-enum` or add the build flag to `package.json` .

JonoPrest:
Thanks very much @CPunisher!

PR #2316, opened 2024-10-18
feat(napi-derive): support string union enum

CPunisher:
closes https://github.com/napi-rs/napi-rs/issues/2246

The string union enum is an extension of string enum. When it is enabled, a default `#[napi(string_enum)]` is also enabled automatically. Users can also specify the case with `#[napi(string_enum_union, string_enum = "xxx")]`.

```rs
#[napi(string_union_enum)]
enum Kind {
  Duck,
  Dog,
  Cat,
}
```
⬇️⬇️⬇️⬇️⬇️⬇️⬇️⬇️⬇️
```ts
export type Kind = 
  | "Duck"
  | "Dog"
  | "Cat"
```


Brooooooklyn:
@CPunisher I remember the behavior of `#[string_enum]` under the `—no-const-enum` build flag is just like this?

CPunisher:
> @CPunisher I remember the behavior of `#[string_enum]` under the `—no-const-enum` build flag is just like this?

Oh, sorry, I did not notice.

Travispersson:
Hey, sorry for chiming in! I'm testing this on the latest 2.x version.
```rust
#[napi(string_enum)]
pub enum Kind {
  Duck,
  Dog,
  Cat,
}
```
Running this command `napi build --platform --release --no-const-enum` results in the following output which is no longer a const enum:
```ts
// string enum
export enum Kind {
  Duck = 'Duck',
  Dog = 'Dog',
  Cat = 'Cat'
}
```
The expected output as described in this issue is not what's happening on 2.x
```ts
// string literal union
export type Kind = 
  | "Duck"
  | "Dog"
  | "Cat"
```

It does work however in ^3.0.0-alpha.52 likely due to this change https://github.com/napi-rs/napi-rs/blob/301388607a37c334713978a84c48ad574e7ed99d/cli/CHANGELOG.md?plain=1#L92 

Would still be nice to have it work in 2.x as well if possible?

JonoPrest:
Just to add here the issue which I've now close: https://github.com/napi-rs/napi-rs/issues/2246

A string literal union would be the most idiomatically preferred type output for our use case. Especially if we could control the casing of the strings etc.

Perhaps the first best step is just to have better support/docs for custom trait implementations.

PR #2388, opened 2024-12-03
fix(napi): delete error reference with raw Error

underfin:
The rolldown `buildEnd` hook argument need to `JsError` and also the `BindingOutputs` need it, here using `JsError::Clone` to do it, but it segment fault at `napi_delete_reference`. So here only delete reference for raw `JsError`.


Boshen:
Can you link the rolldown PR for full context? We can't determine the root issue with just this PR.

underfin:
> Can you link the rolldown PR for full context? We can't determine the root issue with just this PR.

https://github.com/napi-rs/napi-rs/pull/2388#issuecomment-2526719685, You can see it. The `buildEnd` argument and `BindingOutputs#errors` need to  `JsError` to reflect the raw js errors.

Brooooooklyn (approved):

underfin:
@Brooooooklyn Could you publish a new version for this.

Issue #2468, opened 2025-02-15
Allow specifying the library path with an env variable


sigmaSd:
This is sometimes useful, for example I want to package an app that uses napi-rs  as flatpak, with this I can fetch the library at packaging time and have granular control on how to use it, for example I'm compiling an app with deno, this helps me to bundle the library

If this is ok , it should be a simple change, just adding a check for an env variable at the top of this function
https://github.com/napi-rs/napi-rs/blob/b1f8966fa7ed1095daf892cf781b722fb27bf416/examples/napi/index.cjs#L62

sigmaSd:
This will allow for my current app to specify https://www.npmjs.com/package/@slint-ui/slint-ui-binary-linux-x64-gnu location, which allow me to control how to embed it

PR #2471, opened 2025-02-17
feat(cli): allow specifing native library path via env variable

sigmaSd:
fix https://github.com/napi-rs/napi-rs/issues/2468


sigmaSd:
@Brooooooklyn will this end up in 2.x release ?it's not a breaking change so it would be great if that's possible

PR #2485, opened 2025-02-27
feat: class wrapped value with TaggedObject

SyMind:


SyMind:
@Brooooooklyn I believe all test cases have been successfully passed. However, I removed the ClassWithLifetime test case because TypeId::of<T>() only accepts types with a static lifetime.

PR #2511, opened 2025-03-15
fix(cli): use `new URL` instead of `?url` for browser wasm fetching

hi-ogawa:
Related https://github.com/vitejs/vite/issues/19639, https://github.com/oxc-project/oxc/issues/9783

I was testing napi-rs oxc-parser package on browser (see sample app in https://github.com/oxc-project/oxc/issues/9783) and noticed that `?url` is causing some issues on Vite. Using `new URL` works around the issue and also I think it's better to avoid `?url` since it requires a special bundler side convention.


Brooooooklyn (approved):

hi-ogawa:
As I wrote in https://github.com/napi-rs/napi-rs/issues/2517#issue-2923521577, this change would require Vite users to add `optimizeDeps.exclude` for napi-rs browser package, but do you think that's okay?

PR #2028, opened 2024-04-08
feat(target): add support for s390x-unknown-linux-gnu

sapphi-red:
This PR adds support for s390x. Built on top of #2023.

PR #2524, opened 2025-03-18
fix: backport BigInt changes from #2028 & #2356

Wodann:
This fixes issues with marshalling of a `BigInt` to Rust native types.


Wodann:
The failing tests are a result of the build being broken on `napi2`. I'm not sure whether a fix is already in progress?

adespawn:
I know I'm a bit late here, but is there a reason changes without updated tests got backported?

PR #2532, opened 2025-03-29
nwjs support and nwjs example

MikalDev:
Added support for nwjs node library using the same function testing method used for gnu windows, which also defaults to the node and electron methods, before searching for the nwjs node library.

An example of nw.js was also added as an example for testing.


MikalDev:
Unverified signature, removing PR

PR #2605, opened 2025-05-05
fix: Fix path escaping issue in metadata parser

inakineitor:
The way the `cargo metadata` command is currently run is prone to failure because the path is not escaped. The best way to handle this is to pass the arguments directly as separate to the `cargo` process through `spawnSync` so there is no need to rely on manual escaping.

Please let me know if there is anything else I should edit before this can be merged!

I am currently using NAPI v3 so I can also put in the work to get it merged to that version following your guidance.

Here is an example of what happens when the project is run in a path that has a space in the name of a folder.
image


inakineitor:
@Brooooooklyn Do you know if these two tests are failing from the changes or if they are flaky?

PR #2665, opened 2025-05-26
Patch napi-cli: Add freebsd arm64 support

junderw:
This support is currently on 3.0.0 alpha.

We would like to have it on the stable version of napi-cli.

Thank you.


Brooooooklyn:
@junderw 3.0.0 will release very soon, you can safely upgrade to alpha version now

PR #2681, opened 2025-06-02
perf(napi): PersistedPerInstanceHashMap can be thread local

SyMind:
Performance analysis revealed unexpected overhead in `napi::bindgen_runtime::module_register::PersistedPerInstanceHashMap<K,V>::borrow_mut`.

Optimizations

1. Lock mechanism refinement for PersistedPerInstanceHashMap
Replaced internal `RwLock` with `RefCell` since this struct doesn't require Send + Sync implementation, eliminating thread synchronization overhead.

2. Thread-local initialization
Initialized the following global variables via `thread_local!` macro for thread safety:
    - REGISTERED_CLASSES
    - FN_REGISTER_MAP
    - THREADS_CAN_ACCESS_ENV

3. Cross-thread sharing enhancement
Refactored `ModuleClassProperty` into a struct guarded by `RwLock` to enable safe sharing between JS main thread and worker threads via `MODULE_CLASS_PROPERTIES`.

image

### DIFF
--- a/crates/napi/src/bindgen_runtime/js_values/arraybuffer.rs
+++ b/crates/napi/src/bindgen_runtime/js_values/arraybuffer.rs
@@ -369,8 +369,7 @@ macro_rules! impl_typed_array {
             if CUSTOM_GC_TSFN_DESTROYED.load(Ordering::SeqCst) {
               return;
             }
-            if !THREADS_CAN_ACCESS_ENV.borrow_mut(|m| m.get(&std::thread::current().id()).is_some())
-            {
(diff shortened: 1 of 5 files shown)


Brooooooklyn (approved):

Brooooooklyn:
Notes to look back on:

I was trying to avoid `thread_local!` usage because of I wasn't figure out why https://github.com/napi-rs/napi-rs/blob/napi%403.0.0-beta.4/.cargo/config.toml#L4

We can use `thread_local!` safely now

PR #2698, opened 2025-06-07
fix: (wasm-runtime) fs race condition

lyonbot:
the `encodeValue` is expensive, and it seems Chrome may do out-of-order execution, causing the Worker fails to get the length and content.

The `Atomics.store` blocks any other read or write operation against the array.
Updating the status code after writing data might be feasible. 

POC: https://stackblitz.com/edit/stackblitz-starters-nmf8pm43


Brooooooklyn:
bugbot run

Brooooooklyn:
@toyobayashi, could you please take a look?

Issue #2708, opened 2025-06-11
Mem leak in define_properties


rracariu:
This is a continuation of #2696 - observed that for deno and bun there's a memory leak caused by this:

https://github.com/napi-rs/napi-rs/blob/6f74589a252c4c9ca344f7b18a066906260c704a/crates/napi/src/bindgen_runtime/js_values/object.rs#L326

That allocation adds 8 bytes that are not freed. Strangely, on nodejs the leak is no present. Deno and Bun are leaking however.

The fix here is to no allocate if `closures.len()` is 0 and also not add finalizers.

Brooooooklyn:
@rracariu I compared Deno's source code, and I suspect this might be a bug in Deno's implementation

The napi_add_finalizer in deno: https://github.com/denoland/deno/blob/v2.3.5/ext/napi/js_native_api.rs#L3496 created a Weak Reference on Object, and it calls with `v8::Weak::with_finalizer` https://github.com/denoland/deno/blob/v2.3.5/ext/napi/js_native_api.rs#L119. And in `rusty_v8` source code here: https://github.com/denoland/rusty_v8/blob/v137.2.0/src/handle.rs#L623-L627 

> There is no guarantee as to *when* or even *if* the finalization callback
> will be invoked. The invocation is performed solely on a best effort
> basis. GC-based finalization should *not* be relied upon for any critical
> form of resource management! Consider using
> [`Self::with_guaranteed_finalizer`] instead.

PR #2711, opened 2025-06-11
fix(napi): avoid add finalizer if properties is empty

Brooooooklyn:
Maybe can resolve https://github.com/napi-rs/napi-rs/issues/2708, also boost some perf

### DIFF
--- a/crates/napi/src/bindgen_runtime/js_values/object.rs
+++ b/crates/napi/src/bindgen_runtime/js_values/object.rs
@@ -318,23 +318,28 @@ pub trait JsObjectValue<'env>: JsValue<'env> {
     let env = self.value().env;
     #[cfg(feature = "napi5")]
     {
-      let mut closures = properties_iter
-        .clone()
-        .map(|p| p.data)
-        .filter(|data| !data.is_null())
-        .collect::<Vec<*mut std::ffi::c_void>>();
-      let len = Box::into_raw(Box::new(closures.len()));
-      check_status!(unsafe {
-        sys::napi_add_finalizer(
-          env,
-          self.value().value,
-          closures.as_mut_ptr().cast(),
-          Some(finalize_closures),
-          len.cast(),
-          ptr::null_mut(),
-        )
-      })?;
-      std::mem::forget(closures);
+      if !properties.is_empty() {
+        let mut closures = properties_iter
+          .clone()
+          .map(|p| p.data)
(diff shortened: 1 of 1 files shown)


Brooooooklyn:
* **#2711** <a href="https://app.graphite.dev/github/pr/napi-rs/napi-rs/2711?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/napi-rs/napi-rs/2711?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

crates/napi/src/bindgen_runtime/js_values/object.rs
```diff
@@ -318,23 +318,28 @@ pub trait JsObjectValue<'env>: JsValue<'env> {
...
-        )
-      })?;
-      std::mem::forget(closures);
+      if !properties.is_empty() {
+        let mut closures = properties_iter
+          .clone()
+          .map(|p| p.data)
+          .filter(|data| !data.is_null())
+          .collect::<Vec<*mut std::ffi::c_void>>();
+        let len = Box::into_raw(Box::new(closures.len()));
```
rracariu, line 327: The problem I noticed is that if you allocate this empty box there is still a leak. My try fix was to avoid this allocation altogether and also not add the finalizers if closures was empty.

Is there a requirement to add finalization regardless if you have closures or not?

Brooooooklyn: I have `if !properties.is_empty() { }` here, I believe it can avoid the empty Box allocation and useless `napi_add_finalizer`

rracariu: Yes, but properties is not empty. The `closures` vec is empty, because the properties added are plain `name: value ` properties.

So my question is if you have properties, but not closures for them, if you still need to add the finalizers...

My fix is just 

```
if closures.len() > 0 {
 // alloc box
 // add finalizer
}

Brooooooklyn: Oh, my bad, will create another pr to fix it

PR #2721, opened 2025-06-22
Added support for ppc64le

mgiessing:
This PR adds support for ppc64le images


Brooooooklyn:
@mgiessing We now have a new cross-compilation toolchain that supports ppc64le architecture without requiring `Docker`, will release soon

PR #2738, opened 2025-06-24
fix(deps): `@octokit/rest`, `@std/toml` and `lodash-es` should be dependencies

JounQin:
also clean up unused devDependencies


JounQin:
close in favor of #2735

PR #2742, opened 2025-06-24
fix(napi-derive): single parentheses arg is provided in ThreadsafeFunction

camc314:


examples/napi/index.d.cts
```diff
@@ -539,6 +539,9 @@ export declare function extendsJavascriptError(errorClass: any): void
 export type ExternalLinterLoadPluginCb =
   ((arg: string) => PluginLoadResult)
 
+export type ExternalLinterLoadPluginCb2 =
+  ((string) => PluginLoadResult)
+
```
camc314: current napi impl is wrong, this should've generated:
```suggestion
export type ExternalLinterLoadPluginCb2 =
  ((arg: string) => PluginLoadResult)
```

camc314:
thanks!!

PR #2768, opened 2025-07-07
fix(cli): check `kHandle` instead

JounQin:


CPunisher:
Sorry, I think it's my typo

PR #2770, opened 2025-07-11
perf(napi): Use linkme instead of ctor for module register

quininer:
This is a draft that attempts to use linkme instead of ctor for module register.

It has several advantages
1. Faster, the module exports list is determined at compile time and does not need to be created by ctor at startup
2. Smaller, which avoids create a lot of ctor functions
3. Safer, the execution order of ctor is hard to determine, and many regular codes cannot run in it, which is why it is required to be unsafe.

The problem with linkme is that it doesn't support wasm, but that doesn't seem to affect us.

I also keep old ctor style interface so that old versions of napi-macro can still be used.

### DIFF
--- a/crates/macro/src/lib.rs
+++ b/crates/macro/src/lib.rs
@@ -152,18 +152,15 @@ pub fn module_exports(_attr: TokenStream, input: TokenStream) -> TokenStream {
   };
 
   let register = quote! {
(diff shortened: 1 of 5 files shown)


Brooooooklyn:
It appears there is an issue with cross-compilation in Zig.

quininer:
This is a bit strange, I can't reproduce this locally (Linux and macOS).

quininer:
I'm closing it for now, maybe I'll check it out again when I have time.

PR #2777, opened 2025-07-15
feat(cli): add `buffer` to wasi browser template

CPunisher:
From https://emnapi-docs.vercel.app/reference/list.html#buffer-related we should the Buffer polyfill is needed in browser for Buffer related api.

I'm not sure whether bundle `buffer` to `@napi-rs/wasm-runtime` is a better choice.

### DIFF
--- a/cli/src/api/build.ts
+++ b/cli/src/api/build.ts
@@ -1056,6 +1056,7 @@ export type TypedArray = Int8Array | Uint8Array | Uint8ClampedArray | Int16Array
           this.config.wasm?.maximumMemory,
           this.config.wasm?.browser?.fs,
           this.config.wasm?.browser?.asyncInit,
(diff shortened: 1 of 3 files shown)


Brooooooklyn:
The `Buffer` is injected into `wasm-runtime` already: https://github.com/napi-rs/napi-rs/blob/main/wasm-runtime/rollup.config.js#L28 need to make it optional too.

CPunisher:
> The `Buffer` is injected into `wasm-runtime` already: https://github.com/napi-rs/napi-rs/blob/main/wasm-runtime/rollup.config.js#L28 need to make it optional too.

I see that `Buffer` is injected for `memfs`. Maybe it's better to make it optional depending on `fs` option in the `package.json`. However, it could be hard because the injection happens in the stage of bundling `@napi-rs/wasm-runtime/fs`.

PR #2785, opened 2025-07-20
feat: remove NapiRaw and NapiValue cfg and add implement for Object

richerfu:
Allow Object to implement the `NapiRaw` and `NapiValue` can help us to remove `JsObject` which is marked as `deprecated`

### DIFF
--- a/crates/napi/src/bindgen_runtime/js_values/object.rs
+++ b/crates/napi/src/bindgen_runtime/js_values/object.rs
@@ -6,7 +6,8 @@ use std::marker::PhantomData;
 use std::ptr;
 
 use crate::{
-  bindgen_prelude::*, check_status, raw_finalize, sys, type_of, Callback, TaggedObject, Value,
+  bindgen_prelude::*, check_status, raw_finalize, sys, type_of, Callback, NapiRaw, NapiValue,
+  TaggedObject, Value,
 };
 #[cfg(feature = "napi5")]
 use crate::{Env, PropertyClosures};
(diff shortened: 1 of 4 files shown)


Brooooooklyn:
@richerfu 

There is `JsValue::raw` on the `Object`:

```rust
use napi::JsValue;

#[napi]
pub fn accept_obj(obj: Object) {
   object.raw(); // access the underlying napi_value
}
```

For the `from_raw`, you can migrate to the `ValidateNapiValue::validate` + `FromNapiValue::from_napi_value`

For the `from_raw_unchecked`, you can migrate to the `FromNapiValue::from_napi_value`

richerfu:
@Brooooooklyn  Got it. And i have a question about it: 

`JsObject` is marked as deprecated, but `CallContext` is not. So if we remove the implemention of `JsObject` later, we can't get object from `CallContext`. What should we do? 
- Remove the `CallContext` 
- Don't support it
- Add `NapiRaw` and `NapiValue` implemention for `Object` which need by `CallContext`
   ```rs
     pub fn this<T: NapiValue>(&self) -> Result<T> {
        unsafe { T::from_raw(self.env.0, self.raw_this) }
      }
    
      pub fn this_unchecked<T: NapiValue>(&self) -> T {
        unsafe { T::from_raw_unchecked(self.env.0, self.raw_this) }
      }
   ```

Brooooooklyn:
`CallContext` and `#[js_function]` should also be deprecated, everything behind the `compat-mode` feature should be deprecated.

PR #2749, opened 2025-06-28
fix(napi): callback should be Fn rather than FnOnce

Brooooooklyn:
In ThreadsafeFunction::call_with_return_value

PR #2790, opened 2025-07-21
feat(napi): callback in tsfn `call_with_return_value` can be `FnOnce`

h-a-n-a:


h-a-n-a:
Looks like this caused the change https://github.com/napi-rs/napi-rs/pull/2749

PR #2812, opened 2025-07-28
test: panic on throw in callback of threadsafe function with CalleeHandled=false

sapphi-red:
This test crashes with `fatal runtime error: failed to initiate panic, error 5`.
Is this expected?

The documents does not say whether it is unsafe to throw inside the callback.
https://napi.rs/docs/concepts/threadsafe-function.en#errorstrategy


sapphi-red:
This behavior is expected.

PR #2813, opened 2025-07-29
fix(napi-derive): generate types for threadsafe_function with WEAK=true correctly

sapphi-red:
The types generated for `ThreadsafeFunction` with `WEAK=true` was not correct. This PR fixes that.


examples/napi/index.d.cts
```diff
@@ -961,6 +961,8 @@ export declare function tsfnThrowFromJs(tsfn: ((err: Error | null, arg: number)
 
 export declare function tsfnThrowFromJsCallbackContainsTsfn(tsfn: ((err: Error | null, arg: number) => Promise<number>)): Promise<void>
 
+export declare function tsfnWeak(tsfn: (() => void)): Promise<void>
```
sapphi-red, line 964: This was
```ts
export declare function tsfnWeak(tsfn: ((err: Error | null, ) => void)): Promise<void>
```

PR #2827, opened 2025-08-04
fix(wasm-runtime): add constructors for nested values in objects

CPunisher:
Add `storeConstructor` and `loadConstructor` functions that recursively add constructors for nested values in objects.

For example, `fs.readdir` returns an array of `DirEntry`. The original code doesn't add `DirEntry` constructor for the elements in the array.

I'm not sure if the code format is correct, you can re-format it as you like.

### DIFF
--- a/wasm-runtime/fs-proxy.cjs
+++ b/wasm-runtime/fs-proxy.cjs
@@ -41,15 +41,35 @@ const encodeValue = (memfs, value, type) => {
       return view
     }
     case 6: {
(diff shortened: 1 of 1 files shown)


Brooooooklyn:
cursor review

Brooooooklyn:
@CPunisher there are some AI reviews, can you take a look?

Brooooooklyn:
cursor review

Brooooooklyn (approved):

PR #2831, opened 2025-08-04
feat(napi): add ScopeGenerator trait

LeMikaelF:
This PR makes it possible to return JS objects from the `next()` method of `Generator`. It's a breaking change, but a very minor one, since the fix is just to add an anonymous lifetime to the `Generator` and a method parameter, like I had to do in the examples. Since this is an unstable feature, I think it should be acceptable.

I tried adding a separate `ScopedGenerator` similar to `ScopedTask`, like what was suggested on the napi-rs Discord, but it wasn't a good fit. `ScopedTask` works with a blanket impl that converts `Task` into a `ScopedTask`, but since `ScopedGenerator::next` needs to receive a `&'a Env` as a parameter, it wasn't possible to cleanly do the conversion.

I also tried tying the lifetime of the yielded object to the lifetime of the struct implementing `Generator`, but this required storing the `Env` in a struct field, and then it wasn't possible to construct the object from JS, because `Env` doesn't implement `ToNapiValue`.

TODO fix type generation: `Fib4` should have a `[Symbol.iterator]` key.


crates/napi/src/bindgen_runtime/iterator.rs
```diff
@@ -285,21 +285,29 @@ extern "C" fn generator_next<T: Generator>(
   if !completed {
     let g = unsafe { Box::leak(Box::from_raw(generator_ptr as *mut T)) };
     let item = if argc == 0 {
-      g.next(None)
+      g.next(
+        // SAFETY: `Env` is long lived
```
LeMikaelF, line 368: I got this hack from `complete_impl` in `async_work.rs`.

Brooooooklyn:
Thanks for your contribution! I want to keep the original `Generator` trait and introduce a new `ScopedGenerator` trait to avoid breaking changes, making `Generator` easier to use since most cases don't require a lifetime.

LeMikaelF:
> Thanks for your contribution! I want to keep the original `Generator` trait and introduce a new `ScopedGenerator` trait to avoid breaking changes, making `Generator` easier to use since most cases don't require a lifetime.

I tried to do that, but no luck:

> I tried adding a separate ScopedGenerator similar to ScopedTask, like what was suggested on the napi-rs Discord, but it wasn't a good fit. ScopedTask works with a blanket impl that converts Task into a ScopedTask, but since ScopedGenerator::next needs to receive a &'a Env as a parameter, it wasn't possible to cleanly do the conversion.

I'm not a Rust expert, so maybe there's some way that I didn't see there, but I did try for a few hours and didn't manage. Do you have a suggestion?

I'll wait for your input before I do any more work on this PR.

Brooooooklyn:
cursor review

PR #2875, opened 2025-08-15
chore(napi): extends the Set types interoperability

lghuahua:
Sometimes it is necessary to iterate over the elements in the collection in the order of insertion, like JS's set


Brooooooklyn:
cursor review

PR #2876, opened 2025-08-15
Added feature libnode

alshdavid:
This adds the ability to use napi-rs with `libnode`.

Usage

```rust
fn main() {
  libnode_sys::load::cdylib("/path/to/libnode.so");
  napi::setup("/path/to/libnode.so");

  napi::napi_module_register(napi::napi_module {
     nm_modname: "example_native_module",
     // ...
  });

  let args = vec!["-e", "console.log(process._linkedBinding('example_native_module'))"];
  // ...
  unsafe { 
    libnode_sys::node_embedding_main(c_args.len() as c_int, c_args.as_ptr())  // Prints '42'
  };
}

// summary
unsafe extern register(env: napi::Env, exports: napi::JsObject) {
  exports.set("meaningOfLife", 42);
}
```


Brooooooklyn:
@alshdavid can you please add a test fixture as well

alshdavid:
Will come back to this after v2 is merged: https://github.com/napi-rs/napi-rs/pull/2880

PR #2897, opened 2025-08-27
cli/templates: auto look for wasi-browser.js as well in the generated index.js

dikatok:
This is especially useful for local development of wasm binding


cli/src/api/templates/js-binding.ts
```diff
@@ -237,6 +237,15 @@ if (!nativeBinding || process.env.NAPI_RS_FORCE_WASI) {
       }
     }
   }
+  if (!nativeBinding) {
+    try {
+    nativeBinding = require('./${localName}.wasi-browser.js')
```
JounQin:
```suggestion
      nativeBinding = require('./${localName}.wasi-browser.js')
```

PR #2902, opened 2025-09-01
[reupload] cli/templates: auto look for wasi-browser.js as well in the generated index.js

dikatok:
Reupload to add verified tag

This is especially useful for local development of wasm binding


dikatok:
ah, I see a problem with this PR, it works for using wasm in node environment alas looking for `./${localName}.wasi.cjs` but not in browser since `createRequire` is not available

PR #2907, opened 2025-09-04
fix(cli): NAPI_RS_NATIVE_LIBRARY_PATH usage

7rulnik:
Currently it's impossible to override binary using `NAPI_RS_NATIVE_LIBRARY_PATH` because `requireNative` function doesn't return value for it. It'sets to `nativeBinding` but later we set it function result `nativeBinding = requireNative()` which leads to always undefined in this case.


7rulnik:
@Brooooooklyn sorry for ping but could you clarify how release works? I mean it did not get into https://github.com/napi-rs/napi-rs/commit/2cb9c17e1ca9333eba2c8a853f482f8efbea5fdd

PR #2931, opened 2025-09-19
chore(cli): replace find-up with empathic

bluwy:
Replace `find-up` (5 deps) with `empathic` (0 deps) for searching directories upwards. This reduces the dependencies installed for `@napi-rs/cli`

The source code for empathic's find function can be found [here](https://github.com/lukeed/empathic/blob/f5f944a06f9a411168a0073c3fe14e517031d098/src/find.ts#L77-L88).

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -68,8 +68,8 @@
     "colorette": "^2.0.20",
     "debug": "^4.4.1",
     "emnapi": "^1.5.0",
+    "empathic": "^2.0.0",
     "es-toolkit": "^1.39.10",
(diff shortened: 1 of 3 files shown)


bradzacher:
It's worth calling out that `find-up` is also an ESM-only package which meant that `@napi-rs/cli` only works on Node v20.19 and newer (even though the package is listed as working on `>= 16`)
https://github.com/napi-rs/napi-rs/blob/cdccff74cc5b940fa3cf9f6202f22651832152a0/cli/package.json#L9-L11

So this switch fixes that regression!

PR #2996, opened 2025-11-05
feat(sys): use libloading to load napi symbols at runtime on all platform

### LINKED ISSUE #2968: Stop tweaking linker configurations
## Background:

We added more and more linker configs to fix napi failing to run cargo tests caused by missing symbols.

This is due to ... @Brooooooklyn please fill in

In oxc, we now have 

<details>

```
# Fix napi breaking in test environment <https://github.com/napi-rs/napi-rs/issues/1005#issuecomment-1011034770>
# To be able to run unit tests on macOS, support compilation to 'x86_64-apple-darwin'.
[target.'cfg(target_vendor = "apple")']
rustflags = ["-C", "link-args=-Wl,-undefined,dynamic_lookup,-no_fixup_chains"]

[target.x86_64-pc-windows-msvc]
rustflags = ["-C", "target-feature=+crt-static"]

[target.i686-pc-windows-msvc]
rustflags = ["-C", "target-feature=+crt-static"]

[target.aarch64-pc-windows-msvc]
rustflags = ["-C", "target-feature=+crt-static"]

[target.'cfg(target_os = "linux")']
rustflags = ["-C", "linker-features=-lld", "-C", "link-args=-Wl,--warn-unresolved-symbols"]

# To be able to run unit tests on Windows, support compilation to 'x86_64-pc-windows-msvc'.
# Use Hybrid CRT to reduce the size of the binary (Coming by default with Windows 10 and later versions).
[target.'cfg(target_os = "windows")']
rustflags = [
  "-C",
  "link-args=/FORCE",
  "-C",
  "link-args=/NODEFAULTLIB:libucrt.lib",
  "-C",
  "link-args=/DEFAULTLIB:ucrt.lib",
]

# Addresses a potential segmentation fault issue that occurs when
# running napi-rs within a Node.js worker thread on GNU/Linux systems.
# See https://x.com/Brooooook_lyn/status/1895848334692401270
[target.'cfg(target_env = "gnu")']
rustflags = ["-C", "link-args=-Wl,-z,nodelete"]
```

</details>

which is kind of absurd, and you may notice that we need to switch the back to lld on linux.

---

## Proposal

Goal is to remove all the linker configurations.

Generate stub symbols during tests and let it crash during runtime instead of linking time.

Boshen:
fixes #2968

oxc PR: https://github.com/oxc-project/oxc/pull/15323

```diff
- [target.'cfg(target_vendor = "apple")']
- rustflags = ["-C", "link-args=-Wl,-undefined,dynamic_lookup,-no_fixup_chains"]

- [target.'cfg(target_os = "linux")']
- rustflags = ["-C", "linker-features=-lld", "-C", "link-args=-Wl,--warn-unresolved-symbols"]

-   "-C",
-   "link-args=/FORCE",

- [target.'cfg(target_env = "gnu")']
- rustflags = ["-C", "link-args=-Wl,-z,nodelete"]
```

Also fixes random breakages for `cargo test --all-features` when napi is involved.

See https://github.com/rust-lang/rust/issues/147707#issuecomment-3404375787 for the background, **relying on --warn-unresolved-symbols (ELF) or -undefined dynamic_lookup (Mach-O) to get weak symbols is unsound**

### DIFF
--- a/.github/workflows/test-release.yaml
+++ b/.github/workflows/test-release.yaml
@@ -403,13 +403,19 @@ jobs:
           node-version: 22
           cache: 'yarn'
 
+      - name: Install
+        uses: dtolnay/rust-toolchain@stable
+        if: matrix.settings.host != 'windows-11-arm'
+        with:
+          targets: ${{ matrix.settings.target }}
+
       - name: Install dependencies
         run: yarn install --immutable --mode=skip-build
 
       - name: Install ziglang
         uses: mlugg/setup-zig@v2
         with:
-          version: 0.15.1
+          version: 0.15.2
 
       - name: Install cargo-zigbuild
         uses: taiki-e/install-action@v2


crates/build/src/lib.rs
```diff
@@ -36,6 +39,14 @@ pub fn setup() {
         windows::setup_gnu();
       }
     }
+    Ok("linux") => {
+      if target_env == "gnu" && target_arch == "x86_64" {
+        if let Ok(version) = rustc_version::version() {
+          // https://blog.rust-lang.org/2025/09/01/rust-lld-on-1.90.0-stable/
+          if version.major >= 1 && version.minor >= 90 {}
```
JounQin: Won't work for version like `2.0`


```suggestion
          if (version.major > 1) || (version.major == 1 && version.minor >= 90) {}
```

Issue #2970, opened 2025-10-22
regression: Node.JS process doesn't exit on Windows


barjin:
The Node.JS process using a `napi-rs`-built package hangs indefinitely when it should exit on Windows. At this stage, the process doesn't accept any signals (CTRL+C etc.) and has to be killed.

No sightings of such behavior on Linux / MacOS.

I bisected the `napi-rs` history - this error is not present in https://github.com/napi-rs/napi-rs/commit/eef822142151486e0835c47419c16afc6ff880c1, but reproduces with every version after https://github.com/napi-rs/napi-rs/commit/913e42f47c2303c68f76016396a85759e7cf269f (so likely introduced in #2850).

```typescript
import { Impit } from "./index.wrapper.js";

const impit = new Impit();
const response = await impit.fetch("http://localhost:3000/");

console.log(await response.text());
// Logs the response text, but hangs indefinitely afterwards
// Removing all async calls => doesn't hang and exits as expected.
```

Unfortunately, my knowledge of NAPI internals is limited. Feel free to let me know if I can provide any more information. Thank you for this project!

Original report at https://github.com/apify/impit/issues/290

Brooooooklyn:
I can't reproduce this with the latest NAPI-RS:

<img width="2896" height="1644" alt="Image" src="https://github.com/user-attachments/assets/3f8582f7-f30b-4c75-852b-cf412c4c29b0" />

Repeat the test 100 times, and no hang was observed.

<img width="3820" height="979" alt="Image" src="https://github.com/user-attachments/assets/65fc6c55-1465-473b-a515-e4aeabcfdf15" />

barjin:
Thank you for looking into this, @Brooooooklyn. This is interesting — this exact script consistently caused the Node.js process to hang for me.

Apologies if I’m being too direct, but are you testing this on Windows? The terminal theme in the second screenshot looks quite Linux-like. For reference, aside from the original issue author, I was also able to reproduce this on a Google Cloud Windows Server 2025 instance via RDP.

Brooooooklyn:
I tested on my Windows 11 PC, with powershell and warp terminal

barjin:
[...]

PR #2850, opened 2025-08-08
fix(napi): ensure tokio runtime is initialized for dlopen

Brooooooklyn:
## Summary
- Fixes async functions panic when using nodejs dlopen (#2847)
- Ensures tokio runtime is properly initialized before accessing it

## Problem
When modules are loaded via `dlopen()` instead of `require()`, async functions would panic with:
```
thread '<unnamed>' panicked at /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/napi-3.2.0/src/tokio_runtime.rs:113:6:
Access tokio runtime failed in spawn
```

## Solution
Added `start_async_runtime()` calls to all runtime access functions:
- `spawn()`
- `block_on()`
- `spawn_blocking()`
- `within_runtime_if_available()`

This ensures the runtime is properly initialized even when the module is loaded via `dlopen` instead of the normal `require` path.

## Test plan
- [x] Existing tests pass
- [x] Manually tested with dlopen loading a module with async functions
- [x] Verified async functions (`readFileAsync`, `asyncMultiTwo`) work correctly

Fixes #2847

🤖 Generated with [Claude Code](https://claude.ai/code)

Issue #2847, opened 2025-08-08
Async functions panic when using nodejs dlopen.


omasakun:
When loading a native module created with napi-rs v3 using `dlopen`, asynchronous function calls panic.
This problem did not occur with napi v2.

This issue can be avoided by using `require` instead of `dlopen`, so this issue does not need to be resolved, but is this the expected behavior?

## Reproduction

1. Create a napi project with `napi new` and `cargo add napi -F async`
2. Add an async function
    ```rs
    // lib.rs
    
    use napi_derive::napi;
    
    #[napi]
    pub async fn hello() -> u32 {
      42
    }
    ```
3. Create a loader file:
    ```js
    // loader.mjs
    
    import { join } from "node:path";
    import { dlopen } from "node:process";
    import { fileURLToPath } from "node:url";
    
    const nativeBinding = { exports: {} };
    dlopen(nativeBinding, join(fileURLToPath(import.meta.url), "..", `module.linux-x64-gnu.node`));
    
    globalThis.hello = nativeBinding.exports.hello;
    ```
4. Build & Start REPL: `npm build && node --import ./loader.mjs`
5. Call `hello()`

<details>

<summary>Result</summary>

```
thread '<unnamed>' panicked at /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/napi-3.2.0/src/tokio_runtime.rs:113:6:
Access tokio runtime failed in spawn
stack backtrace:
   0:     0x768fb7f16c1a - std::backtrace_rs::backtrace::libunwind::trace::h88deb10bd0145eb8
                               at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/../../backtrace/src/backtrace/libunwind.rs:116:5
   1:     0x768fb7f16c1a - std::backtrace_rs::backtrace::trace_unsynchronized::he1036f5481c14dff
                               at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x768fb7f16c1a - std::sys::backtrace::_print_fmt::hecc345b6e70c4b20
                               at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/sys/backtrace.rs:66:9
[...]

PR #3028, opened 2025-11-21
fix(napi): tokio runtime shutdown at exit

bmjhversteeg:
This is a proposal to fix https://github.com/napi-rs/napi-rs/issues/2970

I found that https://github.com/mmastrac/rust-ctor/issues/304#issuecomment-2571568746 and https://github.com/mmastrac/rust-ctor/wiki/FAQ:-Life%E2%80%90before-and-life%E2%80%90after-main#important-considerations, discourage using `#[dtor]`  to cleanup tokio runtime, which might explain the issue.

Disclaimer: I'm not very familiar with `napi-rs`. I'm not sure why the second `thread_cleanup()` was added in #2850 and whether or not it was an integral part of fixing that particular issue.

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed public re-export of the `ctor` macro from the bindgen runtime module.

* **Refactor**
  * Simplified conditional compilation logic for module registration and thread cleanup.
  * Updated thread cleanup function signature and implementation for improved C ABI compatibility.

### DIFF
--- a/crates/napi/src/bindgen_runtime/mod.rs
+++ b/crates/napi/src/bindgen_runtime/mod.rs
@@ -2,7 +2,6 @@ use std::ffi::c_void;
 use std::rc::Rc;
 
 pub use callback_info::*;
-pub use ctor::ctor;
 pub use env::*;
 pub use iterator::Generator;
 pub use js_values::*;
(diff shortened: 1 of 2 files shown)


Brooooooklyn:
Using `napi_wrap` on `exports` to clean up the tokio runtime could lead to the following issue: https://github.com/napi-rs/napi-rs/issues/2847

Brooooooklyn:
We can use this strategy on the `wasm` target because WebAssembly cannot use `process.dlopen` to load NAPI-RS modules.

PR #3034, opened 2025-11-26
fix(cli): Fix Windows bindings for Electron/MSYS2 and refresh MSYS2 CI

IronBlood:
Please review.

In this PR:

1. Fixed binding usage on Windows platforms.
    * For x86_64 electron, `msvc`, `gnu`, then `gnullvm` will be tried in order.
    * On MSYS2 x86_64 platforms, `gnu` then `gnullvm` will be tried in order.
    * For regular Windows x86_64, the `msvc` varient will be tried.
    * For aarch64 electron, `msvc` then `gnu` will be tried in order.
    * On MSYS2 CLANGARM64, `gnullvm` will be tried.
    * For regular Windows aarch64, the `msvc` varient will be tried.
2. Refactored the `build-and-test-msys2` job so tests are built and executed.
    * All MSYS2 platforms are included.
    * The dependencies `oxc-node` and `rolldown` are built locally using **patches**, with artifacts placed after `yarn install`. This is a temporary workaround and should be removed once the native addons are published.
    * The workspace `@examples/compat-mode` can be built, but the bindings cannot be loaded, so it's excluded from the workflow. (**UPDATE**: see the comment below)
    * The test `values > threadsafe function return Promise and await in Rust` may fail because it sometimes takes longer than 300ms.
3. Refactored the test `unload.spec.js` to support testing on MSYS2 platforms.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved native binding selection on Windows x64 and arm64 to detect alternative toolchains, honor different shared-library suffixes, and more reliably choose ABIs; unload-test binary resolution now adapts to Node build configurations.

* **CI / Tests**
  * Expanded CI to build and integrate local native tooling and patches, broaden the test matrix (including Electron and profile builds), and run a more comprehensive test sequence using locally built artifacts.

### DIFF
--- a/.github/workflows/test-release.yaml
+++ b/.github/workflows/test-release.yaml
@@ -360,7 +360,7 @@ jobs:
           -[target.'cfg(target_os = "windows")']
           -rustflags = [
           +[target.'cfg(all(target_os = "windows", target_env = "msvc"))']
-          +rustfgags = [
+          +rustflags = [
              # Disables linking against the default Universal C Runtime (libucrt.lib). This prevents conflicts if another CRT library is linked manually.
              "-C",
              "link-args=/NODEFAULTLIB:libucrt.lib",


IronBlood:
Hi, sorry I didn't run all the tests in advance. I'm working on it and enabled the workflow `Test & Release` on my GitHub repo. So currently there's no need to approve the workflows for this PR, I'll let you know when it is fixed.

IronBlood:
I have an [experiment commit](https://github.com/IronBlood/napi-rs/commit/7354c697) which [seems working](https://github.com/IronBlood/napi-rs/actions/runs/21661555090/job/62447328268), but I'm not sure whether I should apply the following change to `examples/napi-compat-mode/Cargo.toml`:

```diff
-napi = { path = "../../crates/napi", features = [
+napi = { path = "../../crates/napi", default-features = false, features = [
+  "napi9",
```

PR #3037, opened 2025-11-28
fix(cli): properly load WASM package in JS binding template

lsndr:
## Summary by CodeRabbit

* **Refactor**
  * Made module loading and WebAssembly path selection more deterministic for more reliable startup.
* **Bug Fixes / Behavior**
  * Added support for an environment override that forces the WASI runtime to be used when requested, improving control over runtime selection.

### DIFF
--- a/cli/src/api/templates/js-binding.ts
+++ b/cli/src/api/templates/js-binding.ts
@@ -238,7 +238,7 @@ if (!nativeBinding || process.env.NAPI_RS_FORCE_WASI) {
       wasiBindingError = err
     }
   }
-  if (!nativeBinding) {
+  if (!nativeBinding || process.env.NAPI_RS_FORCE_WASI) {
     try {
       wasiBinding = require('${pkgName}-wasm32-wasi')
       nativeBinding = wasiBinding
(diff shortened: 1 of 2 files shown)


cli/src/api/templates/js-binding.ts
```diff
@@ -238,7 +238,7 @@ if (!nativeBinding || process.env.NAPI_RS_FORCE_WASI) {
       wasiBindingError = err
     }
   }
-  if (!nativeBinding) {
+  if (!nativeBinding || process.env.NAPI_RS_FORCE_WASI) {
```
lsndr, line 241: Otherwise it will not try to load `wasm` from installed package if there is already found native package

cli/src/api/templates/load-wasi-template.ts
```diff
@@ -144,11 +144,7 @@ const __wasmDebugFilePath = __nodePath.join(__dirname, '${wasmFileName}.debug.wa
 if (__nodeFs.existsSync(__wasmDebugFilePath)) {
   __wasmFilePath = __wasmDebugFilePath
 } else if (!__nodeFs.existsSync(__wasmFilePath)) {
-  try {
-    __wasmFilePath = __nodePath.resolve('${packageName}-wasm32-wasi')
-  } catch {
-    throw new Error('Cannot find ${wasmFileName}.wasm file, and ${packageName}-wasm32-wasi package is not installed.')
-  }
+  __wasmFilePath = __nodePath.resolve('node_modules/${packageName}-wasm32-wasi/${wasmFileName}.debug.wasm')
```
lsndr, line 147: try/catch was redundant here, because `_nodePath.resolve` just resolves path. Also  I fixed path to `wasm` binary

PR #3038, opened 2025-11-28
fix(cli): properly load WASM package in JS binding template

lsndr:
## Summary by CodeRabbit

* **New Features**
  * Allow forcing the WASI runtime via an environment variable (including a strict "error" mode).

* **Changes**
  * Preserve and chain loader errors when WASI is forced to improve diagnostics.
  * Improve fallback resolution to locate the WASM runtime file more reliably and surface clearer errors on failure.

---

> [!NOTE]
> Strengthens WASI loading and asset resolution across templates and examples.
> 
> - In `js-binding.ts` and `examples/napi/index.cjs`, attempt WASI loading when `NAPI_RS_FORCE_WASI` is set (even if native binding exists) and preserve/chain `wasiBindingError` causes; add errors to `loadErrors`.
> - In `load-wasi-template.ts` and `examples/napi/example.wasi.cjs`, locate `.wasm` via `require.resolve('<pkg>/<file>.wasm')` instead of `path.resolve`, with clearer errors when the WASM package is missing.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 18fe2c3575d9bdf98a9b12affe5e304a5d8e64f6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/examples/napi/example.wasi.cjs
+++ b/examples/napi/example.wasi.cjs
@@ -39,7 +39,7 @@ if (__nodeFs.existsSync(__wasmDebugFilePath)) {
   __wasmFilePath = __wasmDebugFilePath
 } else if (!__nodeFs.existsSync(__wasmFilePath)) {
   try {
-    __wasmFilePath = __nodePath.resolve('@examples/napi-wasm32-wasi')
+    __wasmFilePath = require.resolve('@examples/napi-wasm32-wasi/example.wasm32-wasi.wasm')
   } catch {
     throw new Error('Cannot find example.wasm32-wasi.wasm file, and @examples/napi-wasm32-wasi package is not installed.')
   }
--- a/examples/napi/index.cjs
+++ b/examples/napi/index.cjs
@@ -536,13 +536,17 @@ if (!nativeBinding || process.env.NAPI_RS_FORCE_WASI) {
       wasiBindingError = err
     }
   }
-  if (!nativeBinding) {
+  if (!nativeBinding || process.env.NAPI_RS_FORCE_WASI) {
     try {
       wasiBinding = require('@examples/napi-wasm32-wasi')
       nativeBinding = wasiBinding
     } catch (err) {
       if (process.env.NAPI_RS_FORCE_WASI) {
-        wasiBindingError.cause = err
+        if (!wasiBindingError) {
+          wasiBindingError = err
+        } else {
+          wasiBindingError.cause = err
+        }
         loadErrors.push(err)
       }
     }


cli/src/api/templates/js-binding.ts
```diff
@@ -238,7 +238,7 @@ if (!nativeBinding || process.env.NAPI_RS_FORCE_WASI) {
       wasiBindingError = err
     }
   }
-  if (!nativeBinding) {
+  if (!nativeBinding || process.env.NAPI_RS_FORCE_WASI) {
```
lsndr, line 241: Otherwise it will not try to load wasm from installed package if there is already found native package

cli/src/api/templates/load-wasi-template.ts
```diff
@@ -144,11 +144,7 @@ const __wasmDebugFilePath = __nodePath.join(__dirname, '${wasmFileName}.debug.wa
 if (__nodeFs.existsSync(__wasmDebugFilePath)) {
   __wasmFilePath = __wasmDebugFilePath
 } else if (!__nodeFs.existsSync(__wasmFilePath)) {
-  try {
-    __wasmFilePath = __nodePath.resolve('${packageName}-wasm32-wasi')
-  } catch {
-    throw new Error('Cannot find ${wasmFileName}.wasm file, and ${packageName}-wasm32-wasi package is not installed.')
-  }
+  __wasmFilePath = __nodePath.resolve('node_modules/${packageName}-wasm32-wasi/${wasmFileName}.debug.wasm')
```
lsndr: try/catch was redundant here, because _nodePath.resolve just resolves path. Also I fixed path to wasm binary

Brooooooklyn: You should always resolve by package name rather than resolve `node_modules/${packageName}`.

lsndr: But `path.resolve` just resolves a path. If I put just `${packageName}` and not a real path to wasm binary, it will be invalid and will always fail.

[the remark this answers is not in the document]
lsndr: Hm. I don't think it makes debugging difficult. It's aligned with other code which resolves path

cli/src/api/templates/load-wasi-template.ts
```diff
@@ -144,11 +144,7 @@ const __wasmDebugFilePath = __nodePath.join(__dirname, '${wasmFileName}.debug.wa
 if (__nodeFs.existsSync(__wasmDebugFilePath)) {
   __wasmFilePath = __wasmDebugFilePath
 } else if (!__nodeFs.existsSync(__wasmFilePath)) {
-  try {
-    __wasmFilePath = __nodePath.resolve('${packageName}-wasm32-wasi')
-  } catch {
-    throw new Error('Cannot find ${wasmFileName}.wasm file, and ${packageName}-wasm32-wasi package is not installed.')
-  }
+  __wasmFilePath = __nodePath.resolve('node_modules/${packageName}-wasm32-wasi/${wasmFileName}.wasm')
```
Brooooooklyn:
```suggestion
  __wasmFilePath = __nodePath.resolve('${packageName}-wasm32-wasi/${wasmFileName}.wasm')
```

In some installation modes of package managers, `${packageName}-wasm32-wasi` may not be under the `node_modules` of the same directory (hoist)

lsndr: You are right. I utilized `require.resolve` which will help find real path to wasm binary

lsndr:
@coderabbitai review

PR #3241, opened 2026-04-16
fix(cli): correctly detect glibc from interpreter, fixes #3223

### LINKED ISSUE #3223: Do not hardcode `/usr/bin/ldd`
https://github.com/napi-rs/napi-rs/blob/6291a534822e89c5025f92af52adcc96cab9a3c9/examples/napi/index.cjs#L26-L32

This makes it very hard to ship appimages with glibc bundled, because the application thinks that the system is musl and attempts to find a `musl.node` file when it is in fact using glibc.

https://github.com/pkgforge-dev/Tutanota-Desktop-AppImage-Enhanced/blob/d7423ea5cd65931501d32bf0892310cc1d46a4a7/make-appimage.sh#L17-L19

c32-4e3-844-41ed9dc5d:
Fixes #3223 , code heavily inspired by https://github.com/lovell/detect-libc/commit/4d6eafecb2bd6c45f8a95822966cf7c08f79d21b

## Summary by CodeRabbit

* **Chores**
  * Enhanced Linux environment detection for improved compatibility across different distributions.


Samueru-sama:
@c32-4e3-844-41ed9dc5d can you explain how this works? 

Because if I'm reading this right, you are reading the binary using `/proc/self/exe` right? 

I believe this will actually make things worse in my case.

We use somethign called [sharun](https://github.com/VHSgunzo/sharun) which executes the binaries using a bundeld dynamic linker using userland-execve. 

This is done to prevent `/proc/self/exe` from being `ld-linux.so`. However `/proc/self/exe` is now a hardlink to sharun with the name of the binary, and this hardlink is actually a static binary, so I don't know what will happen if you attempt to read it.

c32-4e3-844-41ed9dc5d:
I didn't actually "write" the change, I merely adapted the change from https://github.com/lovell/detect-libc/issues/29#issuecomment-3282128975 which correctly detects glibc in my case, happy to have somebody smarter fix the issue for good.

cli/src/api/templates/js-binding.ts
```diff
@@ -87,6 +90,63 @@ const isMusl = () => {
   return musl
 }
 
+const isMuslFromInterpreter = () => {
+  try {
+    const SELF_PATH = '/proc/self/exe';
+    const MAX_LENGTH = 2048;
+    const fd = openSync(path, 'r');
```
PascalHonegger, line 97: This should probably be `SELF_PATH` both here and in the example?

```suggestion
    const fd = openSync(SELF_PATH, 'r');
```

Samueru-sama:
> I didn't actually "write" the change, I merely adapted the change from [lovell/detect-libc#29 (comment)](https://github.com/lovell/detect-libc/issues/29#issuecomment-3282128975) which correctly detects glibc in my case, happy to have somebody smarter fix the issue for good.

Yeah no, this will actually make things worse for us. 

If it is not possible to reliably tell the dynamic linker without having to read `/proc/self/exe` it would be good to have some env var to tell the force 'hey this is actually musl or glibc'.

For example apps that hardcode `/usr/bin/ldd` are easily fixable since that can be patched with sed, and for applications that are mimified or compressed in some sort, we can still override calls to `/usr/bin/ldd` by preloading a library that overrides paths. 

Here the logic is very different and I don't think we could easily fix it by preloading a library.

PR #3250, opened 2026-04-19
fix(wasm-runtime): preserve Buffer return values in fs-proxy

### LINKED ISSUE #3249: wasm-runtime/fs-proxy: Buffer return values are decoded as plain objects instead of Buffer
## Describe the bug

`@napi-rs/wasm-runtime1 does not preserve Buffer return values across `createOnMessage()` / `createFsProxy()`.

When a proxied fs method returns a `Buffer` such as `readFileSync(path)` without an encoding, the value comes back as a plain object like `{ type: 'Buffer', data: [...] }` instead of a real Buffer. That makes proxied fs behavior diverge from Node’s `fs` API.

## Steps to reproduce

Create the following file in project root
```js
import * as memfs from 'memfs'
import { createFsProxy, createOnMessage } from './wasm-runtime/fs-proxy.js'

const realFs = memfs.createFsFromVolume(
  memfs.Volume.fromJSON({ '/buf.bin': 'abc' }),
)

const onMessage = createOnMessage(realFs)
globalThis.postMessage = (data) => onMessage({ data })

const fs = createFsProxy(memfs)
const value = fs.readFileSync('/buf.bin')

console.log(value)
console.log(value.constructor.name)
console.log(Buffer.isBuffer(value))
```

#### Observed behavior

```console
{ type: 'Buffer', data: [97, 98, 99] }
Buffer
true
```

#### Expected behavior

```console
<Buffer 61 62 63>

```

trivikr:
Fixes: https://github.com/napi-rs/napi-rs/issues/3249

## Summary by CodeRabbit

* **New Features**
  * Added support for binary buffer values with automatic detection, serialization, and deserialization in file operations. Binary file data is now properly handled and transmitted.

* **Tests**
  * Added validation tests for buffer handling in file read operations to ensure proper return values and data integrity.

### DIFF
--- a/wasm-runtime/fs-proxy.js
+++ b/wasm-runtime/fs-proxy.js
@@ -6,6 +6,7 @@
 const getType = (value) => {
   if (value === undefined) return 0
   if (value === null) return 1
+  if (isBuffer(value)) return 10
   const t = typeof value
   if (t === 'boolean') return 2
   if (t === 'number') return 3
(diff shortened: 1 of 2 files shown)


Brooooooklyn:
@codex review

PR #3252, opened 2026-04-20
fix(napi): fix custom tokio runtime recreation and cleanup hook re-re…

### LINKED ISSUE #3251: User-defined Tokio Runtime Cannot Be Recreated After Shutdown
## Problem Description

The current implementation of custom Tokio runtime management in napi-rs has a critical flaw: **user-defined runtimes are single-use and cannot be recreated after being shutdown**. This violates the documented behavior and causes unexpected fallback to default runtime configurations.

## Current Behavior

1. **One-time consumption**: When a user provides a custom runtime via `create_custom_tokio_runtime()`, it's stored in `USER_DEFINED_RT` and marked with `IS_USER_DEFINED_RT = true`.

2. **Runtime creation**: On first use, `create_runtime()` takes the custom runtime from `USER_DEFINED_RT` (via `take()`) and moves it to `RT`.

3. **Runtime shutdown**: When `shutdown_async_runtime()` is called, the runtime in `RT` is shutdown and discarded.

4. **Failed recreation**: Subsequent calls to `start_async_runtime()` attempt to recreate the runtime via `create_runtime()`, which:
   - Sees `IS_USER_DEFINED_RT = true`
   - Attempts to get runtime from `USER_DEFINED_RT`, but finds `None` (already taken)
   - Falls back to creating a default runtime
   - **User's custom configuration is lost!**

## Expected Behavior

Based on the code comments and reasonable expectations:
- User-defined runtimes should be recreatable with the same configuration after shutdown
- The system should support runtime lifecycle management for scenarios like:
  - Electron renderer process reloads
  - Test suites that need to restart the runtime
  - WASM targets requiring manual runtime management

### Problematic Code Snippets

```napi-rs/crates/napi/src/tokio_runtime.rs#L12-24
if IS_USER_DEFINED_RT.get().copied().unwrap_or(false) {
  // Try to take the user-defined runtime if it's still available
  if let Some(user_defined_rt) = USER_DEFINED_RT
    .get()
    .and_then(|rt| rt.write().ok().and_then(|mut rt| rt.take()))
  {
    return user_defined_rt;
  }
  // If the user-defined runtime was already taken, fall back to creating a default runtime
  // This handles the case where the runtime was shutdown and needs to be restarted
}
```

**Issue**: The comment claims this handles restart, but it actually falls back to default runtime, losing user configuration.

### State Management Flaws

1. **`IS_USER_DEFINED_RT` is never reset**: Once set to `true`, it remains `true` forever
2. **`USER_DEFINED_RT` becomes `None` after first use**: The `take()` operation consumes the runtime
3. **No configuration preservation**: User's runtime configuration (thread count, features, etc.) is not stored

### 1: Store Runtime Configuration Instead of Instance
```rust
static USER_DEFINED_RT_BUILDER: OnceLock<tokio::runtime::Builder> = OnceLock::new();

pub fn create_custom_tokio_runtime(builder: tokio::runtime::Builder) {
    USER_DEFINED_RT_BUILDER.get_or_init(|| builder);
    IS_USER_DEFINED_RT.get_or_init(|| true);
}

fn create_runtime() -> Runtime {
    if IS_USER_DEFINED_RT.get().copied().unwrap_or(false) {
        if let Some(builder) = USER_DEFINED_RT_BUILDER.get() {
            return builder.build().expect("Failed to build custom runtime");
        }
    }
    // Default runtime creation...
}
```

### 2: Reset State After Shutdown
```rust
pub fn shutdown_async_runtime() {
    if let Some(rt) = RT.write().ok().and_then(|mut rt| rt.take()) {
        rt.shutdown_background();
        
        // Reset user-defined runtime state if it was used
        if IS_USER_DEFINED_RT.get().copied().unwrap_or(false) {
            // Clear the flag to use default runtime next time
            // Or provide API to recreate with same config
        }
    }
}
```

## Recommendations

1. **Immediate fix**: At minimum, reset `IS_USER_DEFINED_RT` after shutdown to prevent misleading state
2. **Documentation**: Update documentation to clarify that custom runtimes are single-use unless recreated by user

This issue affects all users who:
1. Use custom Tokio runtime configurations
2. Need runtime restart capability (tests, Electron, WASM)
3. Expect consistent runtime behavior across module lifecycle

wwog:
…gistration

- Add `create_custom_tokio_runtime_factory` API that accepts a `Fn() -> Runtime` closure, enabling correct runtime recreation after shutdown (e.g. Electron renderer reload). The existing one-shot `create_custom_tokio_runtime` API is preserved for backward compatibility with a deprecation note.
- Add `debug_assert!` guards to both APIs to catch misuse when both are called.
- Reset `ENV_CLEANUP_HOOK_ADDED` to `false` inside `thread_cleanup` so the next Node.js env (e.g. after Electron window reload) can re-register the cleanup hook. Previously the process-level flag was never reset, causing the runtime to leak across env recreation.

Fixes #3251 

## Summary by CodeRabbit

* **New Features**
  * Added a factory-based API for supplying custom Tokio runtimes, enabling reliable runtime recreation and restart after shutdown.

* **Bug Fixes**
  * Fixed environment cleanup hook handling so hooks can be re-registered when new Node.js env instances are created.

* **Tests**
  * Increased test runner timeout to allow longer-running tests.

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -111,7 +111,7 @@
     "extensions": {
       "ts": "module"
     },
-    "timeout": "2m",
+    "timeout": "4m",
     "workerThreads": false,
     "files": [
       "**/__tests__/**/*.spec.ts",
--- a/crates/napi/src/bindgen_runtime/module_register.rs
+++ b/crates/napi/src/bindgen_runtime/module_register.rs
@@ -626,6 +626,12 @@ fn create_custom_gc(env: sys::napi_env) {
 ))]
 unsafe extern "C" fn thread_cleanup(_data: *mut std::ffi::c_void) {
   if MODULE_COUNT.fetch_sub(1, Ordering::Relaxed) == 1 {
+    // Reset the flag so the next Node.js env (e.g. Electron renderer reload) can
+    // re-register this hook. The hook is bound to the env instance that registered it;
+    // a new env will not inherit it automatically.
+    if let Ok(mut flag) = ENV_CLEANUP_HOOK_ADDED.write() {
+      *flag = false;
+    }
     crate::tokio_runtime::shutdown_async_runtime();
   }
 }
--- a/crates/napi/src/tokio_runtime.rs
+++ b/crates/napi/src/tokio_runtime.rs
@@ -12,15 +12,19 @@ use crate::{JsDeferred, SendableResolver, Unknown};
 fn create_runtime() -> Runtime {
   // Check if we're supposed to use a user-defined runtime
   if IS_USER_DEFINED_RT.get().copied().unwrap_or(false) {
-    // Try to take the user-defined runtime if it's still available
+    // Factory takes priority: supports recreation after shutdown (e.g. Electron renderer reload)
+    if let Some(factory) = USER_DEFINED_RT_FACTORY.get() {
+      return factory();
+    }
+    // One-shot path: original create_custom_tokio_runtime API, only works on first creation
     if let Some(user_defined_rt) = USER_DEFINED_RT
       .get()
       .and_then(|rt| rt.write().ok().and_then(|mut rt| rt.take()))
     {
       return user_defined_rt;
     }
-    // If the user-defined runtime was already taken, fall back to creating a default runtime
-    // This handles the case where the runtime was shutdown and needs to be restarted
+    // If the user-defined runtime was already taken and no factory is set, fall back to a default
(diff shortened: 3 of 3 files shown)


wwog:
The solution is implemented only under the condition that existing code remains unchanged. Subsequent solutions are subject to the maintainers' decision.

PR #3312, opened 2026-06-05
fix(cli): emit Node 12 compatible CJS binding loader

NullVoxPopuli-ai-agent:
## Problem

The CJS binding loader emitted by the CLI (`createCommonBinding`, shared by `createCjsBinding` and `createEsmBinding`) uses syntax/APIs that are newer than the Node versions consuming packages still declare support for via their `engines` field. For example, all of `@node-rs/*` declare `engines: { "node": ">= 12" }`, but the generated `index.js` cannot even be **loaded** on Node 12:

| generated code | fails on Node 12 because |
| --- | --- |
| `const { readFileSync } = require('node:fs')` | the `node:` scheme in CommonJS `require()` is only supported on Node ≥ 14.18 / 16 → `Cannot find module 'node:fs'` |
| `process.report?.getReport` / `process.config?.variables?.…` | optional chaining (`?.`) landed in Node 14 → Node 12 fails to **parse** the module entirely |

So the loader's runtime baseline silently moved to Node ≥ 14, breaking the `>= 12` (and `>= 10`) support those packages advertise.

## Fix

Downlevel the generated loader in `js-binding.ts`:

- `require('node:fs')` → `require('fs')` (resolves identically on every Node version)
- `process.report?.getReport` → `process.report && typeof process.report.getReport === 'function'`
- `process.config?.variables?.…` → explicit `&&` guards

No behavioural change on newer Node; the output is now valid down to ES2019 (Node 12).

## Tests

Adds regression tests in `templates.spec.ts` asserting the generated CJS/ESM loaders parse cleanly and contain no `node:`-scheme `require()`, optional chaining, or nullish coalescing — so this baseline can't silently regress again.

```
✔ createCjsBinding is Node 12 compatible: default
✔ createCjsBinding is Node 12 compatible: with version check
✔ createEsmBinding shares the Node-compatible common loader body
```

## Context

Surfaced while adding Node 12 to the CI matrix in [napi-rs/node-rs#1147](https://github.com/napi-rs/node-rs/pull/1147). That repo currently needs the generated loaders hand-patched; regenerating with this change fixes them at the source.

## Summary by CodeRabbit

* **Tests**
  * Added validation tests to ensure generated native-binding loaders remain parseable and compatible with older Node.js runtimes (no newer JS syntax or platform-specific import forms).

* **Chores**
  * Updated loader templates and examples to use explicit compatibility checks, avoid newer language features and "node:" specifiers, and improve Windows detection.

* **Bug Fixes**
  * Adjusted native/WASI fallback handling and error construction to preserve error causes across load attempts.


NullVoxPopuli:
Hi bots!

Humans, 
if you have any questions, lemme know <3

PR #3365, opened 2026-07-01
perf: optimize object field raw property access

intellild:
## Summary

- Add hidden raw named-property runtime helpers for object field get/set and raw field conversion.
- Update object and structured enum codegen to use `napi_get_named_property` / `napi_set_named_property` for compile-time field names.
- Share object/structured enum field token generation helpers and add example coverage for optional/null semantics and structured enum discriminants.
- This follows up on [web-infra-dev/rspack#14649](https://github.com/web-infra-dev/rspack/pull/14649), which reduced bundle size by 372.00KB.
- Add `object bench` to cover object field and structured enum from-JS conversion paths.

## Why This Reduces Size

For a simple object like:

```rust
#[napi(object)]
pub struct StrictObject {
  pub name: String,
}
```

The generated field read used to inline the full getter and error glue at each field site:

```rust
let name_: String = obj
  .get("name")
  .map_err(|err| napi::bindgen_prelude::decorate_field_error(
    err,
    "StrictObject",
    "name",
  ))?
  .ok_or_else(|| napi::bindgen_prelude::missing_field_error("name"))?;
```

That path calls `Object::get`, which creates a JS string key and then calls `napi_get_property`. The `map_err + ok_or_else` code is also repeated for every generated required field.

With this PR, the expanded code becomes:

```rust
let __obj_field_raw_0 = napi::bindgen_prelude::get_named_property_raw(
  env,
  napi::bindgen_prelude::JsValue::raw(&obj),
  std::ffi::CStr::from_bytes_with_nul_unchecked("name\0".as_bytes()).as_ptr(),
)?;

let name_: String = napi::bindgen_prelude::from_raw_required_field(
  env,
  __obj_field_raw_0,
  "StrictObject",
  "name",
)?;
```

So each generated field now only emits:

- one raw named-property lookup using a compile-time C string
- one call into the shared required/optional field conversion helper

The duplicated field error handling moves into runtime helpers, and the property lookup avoids generating a JS string key for each field access. Optional `use_nullable = false` setters get the same treatment: generated code converts only the `Some(inner)` value and calls `set_named_property_raw`, instead of wrapping the raw object and going through generic `Object::set`.

## Benchmark

`object bench` was run 3 times on `main` (`f72afd58`) and this branch (`21b870d0` at measurement time). Values are average op/s.

| case | main avg op/s | branch avg op/s | delta |
|---|---:|---:|---:|
| required object field | 5,438,641 | 6,930,772 | +27.44% |
| missing optional object fields | 4,726,064 | 6,141,638 | +29.95% |
| nested optional js_name field | 4,102,051 | 5,409,583 | +31.88% |
| structured enum discriminant | 1,646,212 | 1,990,960 | +20.94% |

## Validation

- `cargo build -p napi`
- `cargo build -p napi-derive-backend`
- `yarn workspace @examples/napi build`
- `yarn workspace @examples/napi test` (`225 tests passed`, `1 skipped`)
- `yarn workspace bench build`
- `yarn workspace bench bench object.bench.ts --run`

---
_by OpenAI Codex_

## Summary by CodeRabbit

* **New Features**
  * Added support for nullable/optional struct fields in the N-API example, including new factory functions and exports.
  * Extended structured enum examples with a new optional variant in both normal and lowercase forms.
  * Added benchmark cases for required fields, optional fields, nested optional metadata, and structured enum validation.

* **Bug Fixes**
  * Improved WASI loading behavior so it only activates for explicit force settings, reducing unexpected module load failures.


intellild:
@coderabbitai review

intellild:
@coderabbitai review

Brooooooklyn:
@codex review

Brooooooklyn (approved):

PR #3369, opened 2026-07-02
fix(napi): stop ref exception object in ThreadsafeFunction sync-throw path on wasm targets

### LINKED ISSUE #10075: [Bug]: Crashes with `sent an error! Cannot read properties of undefined (reading 'checkGCAccess')` when resolution failed while running by wasm
### Reproduction link or steps

https://stackblitz.com/github/Flowm/vite-tsconfigpaths-sfc-repro?file=README.md&terminal=build

### What is expected?

The resolution failure error to be shown.

### What is actually happening?

The crash happens.

### System Info

```Shell
Stackblitz

  System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 22.22.3 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.8.2 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
```

### Any additional comments?

found in https://github.com/vitejs/vite/issues/22837
probably related to https://github.com/rolldown/rolldown/issues/6527

shulaoda:
Fixes https://github.com/rolldown/rolldown/issues/10075

## The bug

On wasm targets, when a JS callback invoked through a `WithCallback` ThreadsafeFunction call (`call_async` / `call_async_catch`) throws **synchronously**, the whole process crashes with:

```
worker (tid = N) sent an error! Cannot read properties of undefined (reading 'checkGCAccess')
TypeError: Cannot read properties of undefined (reading 'checkGCAccess')
    at napi_reference_unref (.../@emnapi/core/dist/emnapi-core.cjs.js:5902:23)
```

Real-world impact: any resolution failure in a Vite build running on the wasi build of rolldown (e.g. StackBlitz) crashes instead of showing the build error. Vite's `onLog` handler throws synchronously when it sees `UNRESOLVED_IMPORT`, which is enough to bring the process down.

## Root cause

In `call_js_cb`, the pending-exception branch creates a `napi_ref` to the thrown exception object unconditionally:

```rust
status = unsafe { sys::napi_create_reference(raw_env, exception, 1, &mut error_reference) };
```

The resulting `napi::Error { maybe_raw, maybe_env }` is delivered through the oneshot channel to the Rust future awaiting the call, which usually runs on another thread. When that error is dropped there, `Drop for Error` calls `napi_reference_unref(maybe_env, ...)` from the wrong thread. On wasm each thread has its own emnapi context, so the JS thread's env cannot be resolved from the worker: `envStore.get(env)` returns `undefined` and `undefined.checkGCAccess()` throws, killing the worker that hosts the wasm thread.

This is the same class of bug as #2975. That PR fixed the `From<Unknown> for Error` conversions (which cover async promise rejections), but missed this direct `Error` construction inside `call_js_cb`. That is why an `async` callback that throws is fine while a sync throw still crashes.

## The fix

Mirror #2975: on `target_family = "wasm"`, skip `napi_create_reference` and leave `maybe_raw` null. The error message and stack trace are still captured, because `reason` is filled from `get_error_message_and_stack_trace` either way, and the `cause` chain goes through the already-guarded `From<Unknown>` impl.

One detail: the wasm branch must also set `status = napi_ok`. The exception has already been consumed by `napi_get_and_clear_last_exception`, and leaving `status` as `napi_pending_exception` would make `handle_call_js_cb_status` call `napi_fatal_exception` with an empty exception.

### Behavior change (wasm only)

Same trade-off #2975 already made for the async path: JS receives a recreated error whose message contains the original message and stack trace, instead of the original error object instance (custom properties like `err.code` are not preserved). Native targets are unchanged.

## Tests

- New example fn `tsfn_throw_from_js_catch_drop_in_thread` plus a test that reproduces the real-world pattern: the error from a sync throw is moved to and dropped on another thread. On the wasi build this crashes with the exact `checkGCAccess` error before the fix and passes after it. Passes on native both before and after.
- Existing sync-throw assertions (`Throw from ThreadsafeFunction JavaScript callback`, the `call_async_catch` tests) get `WASI_TEST` branches for the recreated-error shape, following the pattern from #2975.

Verified locally:

- full `values.spec.ts`: 164 passed on native (darwin-arm64) and on wasm32-wasip1-threads
- end to end: a Vite build with a failing import on the wasi build of rolldown (rebuilt against this branch) now reports `Rolldown failed to resolve import ...` instead of crashing, using the reproduction from rolldown#10075 (https://github.com/Flowm/vite-tsconfigpaths-sfc-repro)

## Summary by CodeRabbit

* **New Features**
  * Added a new async example API for handling callback errors across threads.
  * Exposed the new function in the WASI and Node-facing example bindings and TypeScript definitions.

* **Bug Fixes**
  * Improved error handling for JavaScript callback failures on WASM targets.
  * Ensured errors can be safely dropped on another thread without crashing, while preserving the original message.

* **Tests**
  * Updated coverage to validate target-specific error behavior and the new API.


Brooooooklyn:
@codex review

Brooooooklyn (approved):

Issue #3357, opened 2026-06-29
Custom GC Crash with cross thread buffer


jimmyhmiller:
In turbopack we were seeing crashes when enabling our thread workers feature (vs our use of processes before). It was originally believed to be related to (#2555, fixed in nodejs/node#55706) but I was seeing the crashes on versions of node with this fixed and found the true underlying cause.

Here is the crash output

```
#
# Fatal error in , line 0
# Check failed: object_ != kGlobalHandleZapValue.
#
#
#
#FailureMessage Object: 0x172d89f88
----- Native stack trace -----

 1: 0x104af495c node::NodePlatform::GetStackTracePrinter()::$_0::__invoke() [/opt/homebrew/Cel
lar/node/26.3.1/lib/libnode.147.dylib]
 2: 0x10565c2ac V8_Fatal(char const*, ...) [/opt/homebrew/Cellar/node/26.3.1/lib/libnode.147.d
ylib]
 3: 0x105bccd6c v8::internal::GlobalHandles::InvokeFirstPassWeakCallbacks() (.cold.1) [/opt/ho
mebrew/Cellar/node/26.3.1/lib/libnode.147.dylib]
 4: 0x104f26928 v8::internal::GlobalHandles::MakeWeak(unsigned long**) [/opt/homebrew/Cellar/n
ode/26.3.1/lib/libnode.147.dylib]
 5: 0x104a5c404 v8impl::Reference::Unref() [/opt/homebrew/Cellar/node/26.3.1/lib/libnode.147.d
ylib]
 6: 0x104a63668 napi_reference_unref [/opt/homebrew/Cellar/node/26.3.1/lib/libnode.147.dylib]
 7: 0x101d6938c napi::bindgen_runtime::module_register::custom_gc::hbaf1fa38cebb19fe [/Users/j
immyhmiller/Documents/Code/vercel/customgc-repro/v2/customgc_v2.node]
 8: 0x104a774ec v8impl::(anonymous namespace)::ThreadSafeFunction::AsyncCb(uv_async_s*) [/opt/
homebrew/Cellar/node/26.3.1/lib/libnode.147.dylib]
 9: 0x10013a3cc uv__async_io [/opt/homebrew/Cellar/libuv/1.52.1/lib/libuv.1.0.0.dylib]
10: 0x1001491b0 uv__io_poll [/opt/homebrew/Cellar/libuv/1.52.1/lib/libuv.1.0.0.dylib]
[...]

PR #3370, opened 2026-07-02
fix(napi): release Error's exception reference via the custom GC when dropped off-thread.

### LINKED ISSUE #3368: `napi::Error` holding a JS-exception reference is `Send`, but `Drop` deletes its `napi_ref` on whatever thread drops it → V8 GlobalHandles corruption
We've been slowly progressing with our napi-v3 migration (https://github.com/NomicFoundation/edr/pull/1385:) due to issues with drops/TSFNs. `napi-3.10.0` actually solved one of the main classes of issues (https://github.com/napi-rs/napi-rs/issues/3357), which allowed me to pin down the last one with help of Claude (which is an existing TODO in the code). 

I've created a repository that should reliably reproduce the issue (it does for me on WSL at least), see https://github.com/nebasuke/napi-error-send-repro. 

I'll follow up with a PR in a bit. 

Crash output:
```
#
# Fatal error in , line 0
# Check failed: object_ != kGlobalHandleZapValue.
#
#
#
#FailureMessage Object: 0x7e593dbf8890
----- Native stack trace -----

 1: 0x10198f1  [node]
 2: 0x27c4c4b V8_Fatal(char const*, ...) [node]
 3: 0x13e92d5  [node]
 4: 0xf30650 napi_reference_unref [node]
 5: 0x7e5940844e8d  [/home/bas/nomic-claude/napi-error-send-repro/repro.node]
 6: 0x7e594084596c  [/home/bas/nomic-claude/napi-error-send-repro/repro.node]
 7: 0x7e5940845826  [/home/bas/nomic-claude/napi-error-send-repro/repro.node]
 8: 0x7e594084642e  [/home/bas/nomic-claude/napi-error-send-repro/repro.node]
 9: 0x7e594086036f  [/home/bas/nomic-claude/napi-error-send-repro/repro.node]
10: 0x7e594069caa4  [/lib/x86_64-linux-gnu/libc.so.6]
11: 0x7e5940729c6c  [/lib/x86_64-linux-gnu/libc.so.6]
```

# Claude Summary
A `napi::Error` created from a JS exception owns a strong `napi_ref` to the JS error object (`maybe_raw`/`maybe_env`). `Error` is `unsafe impl Send` (`error.rs:102`), and its `Drop` (`error.rs:34-53`) calls `napi_reference_unref` + `napi_delete_reference` on the **current** thread. If the error is dropped off the JS thread — which `Send` invites — this deletes a V8 global handle from a foreign thread, corrupting V8's `GlobalHandles` pool. The corruption is delayed and non-local: the process later crashes at an unrelated handle operation with `Fatal error: Check failed: node->IsInUse()`, SIGSEGV in `v8::internal::GlobalHandles::Destroy`/`Create`, or a use-after-free in the napi TSFN close path at env teardown.

The `Drop` impl already carries a matching `// @TODO: deal with Error created with reference and leave it to drop in 'async fn'` (`error.rs:37`). This issue is that TODO manifesting as real-world memory corruption.

This is the same bug class as #3357 (buffers dropped off the JS thread), which 3.10.0 fixed by routing `Buffer`/`TypedArray` ref recycling through the per-env `CustomGcHandle`. `Error`'s ref has no such routing. Reproduced on napi 3.10.0.

### Where the ref-carrying `Error` is created

1. **ThreadsafeFunction `WithCallback` error path** — when the JS callback throws, `call_js_cb` does `napi_create_reference(raw_env, exception, 1, ...)` and stores it in the `Error` handed to the result callback (`threadsafe_function.rs:852` in 3.10.0). Users of `call_with_return_value` naturally forward that `Result` to the thread that initiated the call — e.g. through an mpsc channel to a worker thread. Dropping the `Err` there is UB.
2. **`Promise<T>` rejection** — the `.catch` handler builds the error via `impl From<Unknown> for Error`, which also does `napi_create_reference(..., 1, ...)` (`error.rs:135-158`), and sends it through a channel to the awaiter. **Any `promise.await` polled off the JS thread (e.g. inside `tokio::spawn` on the napi runtime) receives — and eventually drops — this Error off-thread when the promise rejects.** There is no way for the user to avoid this from safe code: the Error materializes on the awaiting thread already.

### Symptoms observed in production code

Debugged in NomicFoundation/edr#1385 (a Node N-API addon with a tokio runtime; error `Result`s from TSFN callbacks and awaited promises crossed threads):

- Flaky crashes, ~7-10% of test-suite runs on linux-arm64 (musl worst; x86-64 rarely surfaces it — weaker memory ordering and allocator behavior on arm64/musl expose the corruption).
- Faces of one corruption: `Check failed: node->IsInUse()` (V8 CHECK in `GlobalHandles::Destroy` during `~Environment`), SIGSEGV in `GlobalHandles::Destroy`/`Create` at *unrelated* `napi_delete_reference` call sites (e.g. a freshly created `Uint8Array` ref), and a use-after-free in `v8impl::ThreadSafeFunction::CloseHandlesAndMaybeDelete`'s uv close callback during `FreeEnvironment → CleanupHandles`.
- First-chance gdb backtrace naming the culprit (thread is a tokio worker):

  ```
  #0 v8::internal::GlobalHandles::Destroy
  #1 v8impl::Reference::~Reference()
  #2 napi_delete_reference
  #3 napi_sys::...::napi_delete_reference
  #4 napi::error::{impl}::drop (error.rs:46)
  #5 drop_in_place<napi::Error>
  ... (Error boxed into a `Box<dyn std::error::Error + Send + Sync>` error chain,
      dropped on a tokio-rt-worker thread)
  ```

- Crash rate went to **0 in 500 looped suite runs** once the addon stopped letting `napi::Error` cross threads (stringify on the JS thread; `mem::forget` for rejection errors that materialize on tokio threads).

### Reproduction

Standalone repro (no tokio, no TSFN — two functions): https://github.com/nebasuke/napi-error-send-repro. `dropErrorOffThread(err)` wraps `err` via `From<Unknown>` (the promise-rejection code path) and drops the Error on a spawned thread; `churnRefs(obj, n)` creates/deletes references on the JS thread concurrently. **Crashes node 10/10 on plain x86-64 Linux** (napi 3.10.0, Node v24.18.0) with `Fatal error: Check failed: object_ != kGlobalHandleZapValue` in `napi_reference_unref` on the spawned thread, or SIGSEGV in `GlobalHandles::Destroy`/`Create`.

### Fix (implemented, validated)

Route the release in `Error::drop` through the per-env `CustomGcHandle` introduced for #3357, exactly like `Buffer`/`TypedArray` drops: capture the handle when the reference is created (`From<Unknown>`, TSFN `call_js_cb`), release directly when dropped on the owning JS thread, enqueue on the custom-GC TSFN from any other thread, and no-op (safe leak) once the owner env is gone. The `custom_gc` callback deletes only when the refcount reaches zero, so `try_clone`d Errors sharing one reference stay correct (buffer refs are never shared — behavior unchanged). This also resolves the `Drop` `@TODO` and makes `Promise<T>` rejections safe to await from any thread.

Branch: <FIX BRANCH/PR URL>. With the fix, the repro goes from **10/10 crashes to 0/20** ("survived" every run). Also validated against the production reproduction that isolated this (native-arm64 CI stress loop): 0 crashes in 500 suite iterations.

### Environment

- napi: 3.10.0 (also 3.9.x)
- Node: reproduced on 22.x, 24.x, 26.x (Node-version independent — the delete is UB regardless)
- Platform: linux-arm64 (gnu + musl; musl surfaces it most reliably), macOS arm64 occasionally; x86-64 rarely

nebasuke:
Fixes #3368. 

# Claude summary

## Problem

A `napi::Error` created from a JS exception owns a `napi_ref` to the JS error object:

- `From<Unknown>` — the `Promise<T>` rejection path — creates it in the `.catch` handler, and the Error is then received (and dropped) by whichever thread awaits the promise;
- the ThreadsafeFunction JS-throw path (`call_js_cb`) creates it and hands the Error to the result callback, from which callers naturally forward it cross-thread.

`Error` is `Send`, but `Drop` released that reference with `napi_reference_unref`/`napi_delete_reference` on **whatever thread ran it**. Off the owning JS thread that mutates V8's `GlobalHandles` concurrently with the JS thread and corrupts it — `Fatal error: Check failed: object_ != kGlobalHandleZapValue`, `Check failed: node->IsInUse()`, or delayed segfaults in `GlobalHandles::Destroy`/`Create` at unrelated call sites. The old `Drop` carried a `@TODO: deal with Error created with reference` acknowledging this. Same bug class as #3357; the standalone repro in #3368 crashes node 10/10 on plain x86-64 Linux.

## Fix

Route the release through the per-env `CustomGcHandle` from #3360, exactly like `Buffer`/`TypedArray` drops:

- capture the owning env's handle when the reference is created (`From<Unknown>`, TSFN `call_js_cb`);
- in `Drop`: release directly on the owning JS thread; enqueue on the custom-GC TSFN from any other thread (read-lock held across the call); no-op (safe leak) once the env is gone (`aborted`);
- the `custom_gc` callback now deletes only when the refcount reaches zero, so `try_clone`d Errors sharing one reference stay correct. Buffer/TypedArray refs are never shared, so their behavior is unchanged (the removed `debug_assert` was buffer-specific).

## Tests

`worker-thread.spec.ts` gains two tests mirroring the #3357/#3360 coverage, backed by new `examples/napi` exports:

- off-thread drop of `From<Unknown>` Errors (spawned thread) with concurrent JS-thread GlobalHandles churn;
- promise rejections awaited on the async runtime (the Error materializes and drops on a tokio worker);
- `try_clone` siblings dropped on different threads (refcount > 0 path);
- Errors stashed in a worker's thread_local, dropped after env teardown (aborted/leak path).

Without the fix these crash deterministically (`Check failed: object_ != kGlobalHandleZapValue` in `napi_reference_unref` on the foreign thread — verified by reverting `crates/napi/src` to `napi-v3.10.0` and rerunning). With the fix: full `examples/napi` suite passes (223 tests), `cargo fmt`/`clippy` clean across feature combos (default, no-default, `napi9,tokio_rt,compat-mode`).

Also validated downstream: the production crash this was root-caused from (NomicFoundation/edr#1385, flaky arm64 SIGSEGV/`node->IsInUse()` at ~7-10% per suite run) went to 0 crashes in 500 stress-loop iterations with this change.

## Relation to #3369 / #2975

#3369 (opened independently the same day, via rolldown) hits the same off-thread `Error`-ref release on **wasm**, where emnapi fails loudly (`checkGCAccess`) instead of corrupting memory. It follows #2975's approach: skip creating the reference on wasm, at the documented cost of losing the original error object (custom properties like `err.code`). The two PRs compose cleanly: after #3369 there is no ref to release on wasm, so this PR's routing is a no-op there, and this PR fixes the silent corruption on native targets, which #3369 doesn't touch. (Textually they edit adjacent lines in `call_js_cb`; whichever lands second has a trivial rebase.)

Why not extend the skip-the-ref approach to native instead? On native the reference is what lets the original JS error object be rethrown intact (identity and custom properties preserved); #2975/#3369 accepted that loss only because emnapi makes the ref unusable across threads anyway. Routing the release keeps native behavior unchanged while making it thread-safe.

## `try_clone` off-thread (closed after review)

`try_clone` used to call `napi_reference_ref` on the current thread unconditionally; for a live `Error` the refcount stays ≥ 1 so this never touches `GlobalHandles`, but the increment itself is not atomic — off the owning JS thread it races a sibling's concurrent release. Now guarded: `try_clone` returns `Err` off the owning thread (it is already fallible; the previous behavior there was silent UB). Covered by `tryCloneErrorOffThread` in `values.spec.ts`.

## Summary by CodeRabbit

- **New Features**
  - Added example helper APIs for off-thread handling of JS-derived `Error` and promise rejections, including cloning behavior checks, worker-thread stress coverage, and thread-local stash/teardown; exposed via the example module and TypeScript/WASI declarations.
- **Bug Fixes**
  - Improved custom-GC–backed reference cleanup and deletion timing for `Error` references, including safer teardown/read-abort handling. Clarified “reference”-focused error wording and restricted `Error` cloning to the owning thread.
- **Tests**
  - Added worker-thread regression tests and updated TypeScript snapshots to validate off-thread drops, cloning semantics, and teardown behavior.


Brooooooklyn:
@codex review

Brooooooklyn (dismissed):

Brooooooklyn:
@copilot resolve the merge conflicts in this pull request

PR #3388, opened 2026-07-10
fix(cli): require Android NDK for cross-builds

Brooooooklyn:
## Summary

- leave Android-host builds on their native toolchain path
- require `ANDROID_NDK_LATEST_HOME` when targeting Android from another host
- fail before Cargo when the NDK environment is unavailable
- add a focused regression test

## Why

The previous path only warned when `ANDROID_NDK_LATEST_HOME` was missing and then constructed toolchain paths from an undefined value. That deferred the failure and produced a less actionable build error.

This now matches the documented Android flow in https://napi.rs/docs/cross-build#android.

## Validation

- `yarn workspace @napi-rs/cli test src/api/__tests__/android-build.spec.ts` (1 passed)
- full CLI suite on the combined change set (121 passed)
- focused Oxlint and Prettier checks

### DIFF
--- a/cli/src/api/__tests__/android-build.spec.ts
+++ b/cli/src/api/__tests__/android-build.spec.ts
@@ -1,0 +1,78 @@
+import { existsSync } from 'node:fs'
+import { mkdir, rm, writeFile } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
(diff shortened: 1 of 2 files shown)


Brooooooklyn:

PR #3389, opened 2026-07-10
fix(cli): apply selected Node-API version to new projects

Brooooooklyn:
## Summary

- validate that `minNodeApiVersion` is a supported integer
- rewrite the generated Cargo `napi` feature to the selected Node-API level
- preserve unrelated dependency features
- disable the default `napi4` feature for Node-API 1-3 while retaining dynamic symbol loading
- add network-free regression coverage

## Why

`napi new --min-node-api` updated the generated package's Node engine constraint but left Cargo on the template's default `napi4` feature. Projects requesting another Node-API level therefore advertised one minimum while compiling for another.

## Validation

- `yarn workspace @napi-rs/cli test src/api/__tests__/new-node-api-version.spec.ts` (7 passed)
- full CLI suite on the combined change set (121 passed)
- focused Oxlint and Prettier checks
- CLI build

### DIFF
--- a/cli/src/api/__tests__/new-node-api-version.spec.ts
+++ b/cli/src/api/__tests__/new-node-api-version.spec.ts
@@ -1,0 +1,81 @@
+import { mkdtemp, readFile, rm, writeFile } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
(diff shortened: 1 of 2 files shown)


Brooooooklyn:

PR #3391, opened 2026-07-10
docs(napi): refresh CLI build example

Brooooooklyn:
## Summary

- use the valid `name` package.json field
- show the current `@napi-rs/cli` v3 dependency
- use `napi.binaryName` instead of the obsolete `napi.name` field
- clarify that the CLI should be installed as a local development dependency

## Why

The README example used outdated package metadata and CLI configuration, so copying it produced an invalid or misleading setup.

## Validation

- Prettier check


Brooooooklyn:

PR #3415, opened 2026-07-21
perf(napi): outline threadsafe callback invocation

Boshen:
## Summary

- outline type-independent threadsafe callback invocation and exception handling
- keep argument and return conversion in the typed callback shell

In Rolldown, 41 `call_js_cb` instantiations shrink from 51,153 to 43,727 assembly bytes and from 29,948 to 25,162 LLVM lines.

## Validation

- `cargo fmt --all -- --check`
- `cargo check -p napi --all-features`
- `cargo test -p napi --lib --all-features`
- `cargo clippy -p napi --all-features`


Boshen:
Closing because the change reduces generated code but does not reduce the actual linked Rolldown binary size.

PR #3416, opened 2026-07-21
perf(napi): outline object field conversion helpers

Boshen:
## Summary

- keep object field conversion helpers out of generated `FromNapiValue` callers
- reduce duplicated missing-field and field-error handling code

In a same-source Rolldown A/B, total assembly shrinks by 4,118 bytes, LLVM IR by 3,694 lines, and the linked Mach-O `__text` section by 7,360 bytes.

## Validation

- `cargo fmt --all -- --check`
- `cargo check -p napi --all-features`
- `cargo test -p napi --lib --all-features`
- `cargo clippy -p napi --all-features`

### DIFF
--- a/crates/napi/src/bindgen_runtime/js_values/object.rs
+++ b/crates/napi/src/bindgen_runtime/js_values/object.rs
@@ -746,7 +746,7 @@ pub unsafe fn set_named_property_raw(
 }
 
 #[doc(hidden)]
-#[inline(never)]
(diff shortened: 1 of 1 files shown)


Boshen:
Closing this approach: forcing the generic conversion helpers out of line is not an acceptable implementation. Removing the existing inline hint was also measured and produced no code-size change.

PR #3432, opened 2026-07-31
fix(napi-derive): look up constructor by js_name in validate()

### LINKED ISSUE #3427: `Either<>`/`Option<&T>` validation fails for classes whose Rust ident differs from `js_name`
## Bug description

`ValidateNapiValue::validate()`, generated for `#[napi]` classes by `gen_napi_value_map_impl`
in `crates/backend/src/codegen/struct.rs`, looks up the registered JS constructor using the
**Rust struct ident** instead of the class's `js_name`. This works only when the ident equals
`js_name` (the common case), and silently breaks whenever they differ via
`#[napi(js_name = "...")]`.

`validate()` is the mechanism `Either<A, B>` and `Option<&T>` parameters use to probe candidate
types before committing to one. So any function parameter typed `Either<&MyClass, _>` or
`Option<&MyClass>` fails to ever match `MyClass` once `MyClass`'s Rust ident differs from its
`js_name`.

## Root cause

In `crates/backend/src/codegen/struct.rs`:

\`\`\`rust
fn gen_napi_value_map_impl(
  name: &Ident,
  to_napi_val_impl: TokenStream,
  has_lifetime: bool,
) -> TokenStream {
  let name_str = name.to_string();
  ...
  let js_name_str = format!("{name_str}\0");   // <-- built from the Rust ident, NOT js_name
  let validate = quote! {
    unsafe fn validate(...) -> napi::Result<...> {
      if let Some(ctor_ref) = napi::bindgen_prelude::get_class_constructor(#js_name_str) {
        ...
\`\`\`

Despite being named `js_name_str`, this variable is derived from `name.to_string()` (the Rust
ident) — the caller (`impl NapiStruct::gen_napi_value_map_impl`, same file) never passes the
class's real `js_name` into this function at all, even though it's readily available as
`self.js_name` and is already threaded correctly into *every other* codegen path in this file
(`gen_default_ctor`, `gen_to_napi_value_ctor_impl`, `gen_to_napi_value_ctor_impl_for_non_default_constructor_struct`,
`gen_register`). Constructors are only ever registered under `js_name` (see those call sites),
so `get_class_constructor()` here searches under the wrong key and `validate()` always returns
`Err`, causing `Either`/`Option<&T>` to never select this type.

## Minimal repro

\`\`\`rust
#[napi(js_name = "Foo")]
pub struct FooImpl {
  pub value: i32,
}

#[napi]
pub fn takes_either(input: Either<&FooImpl, i32>) -> i32 {
  match input {
    Either::A(f) => f.value,
    Either::B(n) => n,
  }
}
\`\`\`

\`\`\`js
const { Foo, takesEither } = require('./index')
takesEither(new Foo(1))
// Throws: "Failed to get constructor of class `FooImpl`" (or similar InvalidArg),
// even though `new Foo(1)` is a real, valid instance.
```

## Expected behavior

`validate()` should look up the constructor under `js_name` (the name it was actually registered
under), matching every other codegen path in the same file. The Rust ident can stay in use for
cosmetic purposes (`TypeName::type_name()`, error-message text) since those aren't registry
lookups.

## Environment

- `napi` / `napi-derive`: 3.9.1 / 3.5.6 (confirmed present at this version; also reproduced at
  3.12.0 / 3.6.1 with `napi-derive-backend` 6.1.0 — the newer "content-derived type tag" work in
  that release doesn't touch this code path)
- A PR with a minimal fix follows.

### PR DESCRIPTION
**Pull request by Boose101**
## What

`gen_napi_value_map_impl` (in `crates/backend/src/codegen/struct.rs`) builds the
`get_class_constructor()` lookup key for the generated `ValidateNapiValue::validate()`
from the **Rust struct ident** (`name.to_string()`) instead of the class's registered
`js_name`. Every other codegen path in this file already threads `self.js_name` into
that key — this one function was missed.

## Why it's a bug

Constructors are only ever registered under `js_name`, so whenever a class's Rust
ident differs from its `js_name`:

```rust
#[napi(js_name = "Foo")]
pub struct FooImpl { pub value: i32 }
```

`get_class_constructor("FooImpl\0")` searches under the wrong key, `validate()` returns
`Err` unconditionally, and any `Either<&FooImpl, _>` / `Option<&FooImpl>` parameter never
matches a real instance:

```js
const { Foo, takesEither } = require('./index')
takesEither(new Foo(1)) // throws InvalidArg, though `new Foo(1)` is valid
```

When ident == `js_name` (the common case) the key is right by coincidence, which is why
this hasn't surfaced before.

## Fix

Thread the already-available `js_name` into `gen_napi_value_map_impl` and use it for the
`get_class_constructor()` lookup key. The Rust ident stays in use for the cosmetic
identifiers in that function (`type_name()`, error text, the content-derived type tag) —
those aren't registry lookups.

## Testing

I hit this in a downstream addon that gives every `#[napi]` class a suffixed Rust ident
while keeping the JS name unchanged (`FooImpl` → `js_name = "Foo"`). With this change,
`Either` / `Option<&T>` parameters resolve those classes again and the binding's test
suite passes. Happy to add an integration test under `examples/` for a
`js_name`-differs-from-ident + `Either`/`Option<&T>` case if you'd like — just point me at
where you'd prefer it.

Closes #3427

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed constructor resolution for classes whose JavaScript names differ from their Rust identifiers.
  * Improved validation of renamed class instances in strict arguments and union types.
  * Added regression coverage for valid instances, fallback handling, and invalid arguments.
  * Updated browser support to use the current WebAssembly package.


Boose101:
@Brooooooklyn if this could be approved and merged quickly I would appreciate it! Thank you!

Brooooooklyn (dismissed):

Brooooooklyn (dismissed):

Brooooooklyn:
@Boose101 you need to update the test snapshots by running `yarn workspace @examples/napi test -u`

Boose101:
@Brooooooklyn updated test snapshots + added coderabbit suggested changes

Boose101:
@Brooooooklyn What is the roll out timeframe? When should I be able to ingest the new napi-derive-backend crate? And thank you so much for the swift merging!

PR #3434, opened 2026-07-31
fix(backend): keep the namespace when resolving cross-namespace type references

### LINKED ISSUE #3406: Type-def generation keys class members & arg-type aliases by bare `js_name`, ignoring namespace
## Summary

`.d.ts` / `.d.cts` generation keys class members and argument-type aliases by the **bare `js_name`**, ignoring the namespace (`js_mod`). Two `#[napi]` classes that share a `js_name` in **different** namespaces therefore generate incorrect type definitions.

## Repro

```rust
#[napi(namespace = "alpha", js_name = "Client")]
pub struct AlphaClient { pub value: i32 }
#[napi(namespace = "alpha")]
impl AlphaClient {
  #[napi(constructor)]
  pub fn new(value: i32) -> Self { Self { value } }
}

#[napi(namespace = "beta", js_name = "Client")]
pub struct BetaClient { pub value: i32 }
#[napi(namespace = "beta")]
impl BetaClient {
  #[napi(constructor)]
  pub fn new(value: i32) -> Self { Self { value } }
}

// a function taking &AlphaClient, for symptom 2:
#[napi]
pub fn read_alpha(_c: &AlphaClient) -> i32 { 0 }
```

## Symptom 1 — class-member merge collision (constructors/methods)

Generated `.d.cts`:

```ts
export declare namespace alpha {
  export class Client { value: number }          // ❌ constructor missing
}
export declare namespace beta {
  export class Client {
    value: number
    constructor(value: number)
    constructor(value: number)                   // ❌ duplicated
  }
}
```

`alpha.Client`'s constructor was stolen and attached to `beta.Client` (twice).

**Cause:** `preprocessTypeDef` in `cli/src/utils/typegen.ts` builds the class-merge map keyed by the bare name:

- `cli/src/utils/typegen.ts:223` — `const classDefs = new Map<string, TypeDefLine>()`
- `cli/src/utils/typegen.ts:235` — `classDefs.set(def.name, def)` (Struct — the beta struct overwrites the alpha entry)
- `cli/src/utils/typegen.ts:237` / `:243` — `classDefs.get(def.name)` (Extends / Impl — both classes' `impl` records merge onto whichever struct is currently in the map, i.e. beta)

The namespace is already known one line earlier (`cli/src/utils/typegen.ts:226` — `const namespace = def.js_mod ?? TOP_LEVEL_NAMESPACE`) and is used correctly for grouping, but not for the merge key. The `impl` record carries `js_mod` (`crates/backend/src/typegen/struct.rs:153`), so the data is present — the CLI just discards it in the key.

## Symptom 2 — argument-type alias drops the namespace

`read_alpha`'s generated signature references a **dangling, unqualified** class name:

```ts
export declare function read_alpha(c: Client): number   // ❌ no top-level `Client` exists
```

It should be `alpha.Client`.

**Cause:** the Rust-type → TS-name alias map stores the **bare `js_name`**:

- `crates/backend/src/typegen/struct.rs:19` — `add_alias(self.name.to_string(), self.js_name.to_string())`
- map defined at `crates/backend/src/typegen.rs:36-41`, consumed at `crates/backend/src/typegen.rs:636` (`ALIAS.with(|aliases| aliases.borrow().get(rust_ty)...)`)

So `AlphaClient` → `Client` (namespace dropped), producing an unqualified reference.

## Suggested fix

- **Symptom 1:** namespace-qualify the `classDefs` key in `preprocessTypeDef` (e.g. ``${def.js_mod ?? TOP_LEVEL_NAMESPACE}::${def.name}``) at the set/get sites (`cli/src/utils/typegen.ts:235/237/243`). `classDefs` is function-local with no other consumer, so blast radius is limited to the collision case.
- **Symptom 2:** make the alias map namespace-aware (qualify the emitted name, or key by the Rust type and emit `js_mod`-qualified) at `crates/backend/src/typegen/struct.rs:19` / `crates/backend/src/typegen.rs:636`.

Snapshot/regression fixtures (`cli/src/utils/__tests__/__snapshots__/typegen.spec.ts.md`, `cli/src/api/__tests__/build-regressions.spec.ts`) will need regeneration + review.

## Context

Surfaced while adding cross-namespace type-tag test fixtures in #3405; those fixtures were removed there to keep that PR scoped, and this is filed to track the underlying type-def bug separately. The generation code is pre-existing (unmodified by #3405).

Boose101:
Fixes #3406.

The issue's write-up pointed at the `ALIAS` map (`typegen.rs` / `handle_generic_type`), but tracing it further, `CLASS_STRUCTS` is actually what resolves a class reference in a function arg/return type or struct field — it's checked before `ALIAS`/`KNOWN_TYPES` in `handle_type_path`, and it also only stores the bare `js_name` with no namespace. So `AlphaClient` and `BetaClient` (both `js_name = "Client"`, different namespaces) are each retrievable fine by their own Rust identifier, but both resolve to the same unqualified `"Client"` string — which can't be correct for both call sites, and produces the dangling top-level reference described in the issue.

This namespace-qualifies both maps:
- `CLASS_STRUCTS` now stores a small `ClassStructRef { js_name, js_mod }` instead of a bare `String`, and the `handle_type_path` lookup renders it as `{js_mod}.{js_name}` when namespaced. The one place that wants the bare name — a factory method's return type, self-referencing its own class from inside that class's own body — still asks for `.js_name` directly, so that unqualified (and correct) self-reference is unaffected.
- `add_alias` takes the same `js_mod` and qualifies the stored alias, since enums/consts/type-aliases never touch `CLASS_STRUCTS` and resolve through `ALIAS` instead — same bug, same fix, for the paths the issue's repro didn't happen to exercise.

Symptom 1 (the `classDefs` merge collision in `cli/src/utils/typegen.ts`) looks like it's already fixed on `main` — `preprocessTypeDef` keys `classDefs` by `${namespace}\0${def.name}` now, not the bare name — so this PR only covers symptom 2.

Added two regression tests in `crates/backend/src/typegen.rs` covering both paths (`class_reference_across_namespaces_keeps_namespace_qualifier` for `CLASS_STRUCTS`, `non_class_alias_across_namespaces_keeps_namespace_qualifier` for `ALIAS`), plus updated the existing `user_class_named_buffer_precedes_the_builtin_buffer_mapping` test for the new `ClassStructRef` shape.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved generated type definitions for classes, constants, enums, and aliases in namespaced JavaScript modules.
  * Preserved module-qualified class names during type resolution.
  * Improved factory return types by using the correct JavaScript class names.
  * Ensured aliases retain their associated module qualification.
* **Tests**
  * Added coverage for namespaced classes and non-class aliases.

### DIFF
--- a/crates/backend/src/typegen.rs
+++ b/crates/backend/src/typegen.rs
@@ -39,7 +39,11 @@ thread_local! {
   static ALIAS: RefCell<HashMap<String, String>> = Default::default();
 }
 
-fn add_alias(name: String, alias: String) {
+fn add_alias(name: String, alias: String, js_mod: Option<&str>) {
+  let alias = match js_mod {
+    Some(js_mod) => format!("{js_mod}.{alias}"),
+    None => alias,
+  };
   ALIAS.with(|aliases| {
     aliases.borrow_mut().insert(name, alias);
   });
@@ -806,9 +810,11 @@ fn handle_type_path(
     } else if rust_ty == "FnArgs" {
       is_passthrough_type = true;
       Some(args.first().unwrap().to_owned())
-    } else if let Some(t) = crate::typegen::r#struct::CLASS_STRUCTS
-      .with(|c| c.borrow_mut().get(rust_ty.as_str()).cloned())
-    {
+    } else if let Some(t) = crate::typegen::r#struct::CLASS_STRUCTS.with(|c| {
+      c.borrow_mut()
+        .get(rust_ty.as_str())
+        .map(|c| c.qualified_name())
+    }) {
       Some((t, false))
     } else if let Some(&(known_ty, _, _)) = KNOWN_TYPES.get(rust_ty.as_str()) {
       handle_known_type(&rust_ty, known_ty, args, is_return_ty)
@@ -916,8 +922,8 @@ pub fn ty_to_ts_type(
 #[cfg(test)]
 mod tests {
   use super::{
-    escape_json, format_js_property_name, ty_to_ts_type, JSDoc, TypeDef,
-    BUFFER_TYPE_IMPORT_MARKER_BASE, BUFFER_TYPE_IMPORT_SENTINEL,
+    add_alias, escape_json, format_js_property_name, handle_generic_type, ty_to_ts_type, JSDoc,
+    TypeDef, BUFFER_TYPE_IMPORT_MARKER_BASE, BUFFER_TYPE_IMPORT_SENTINEL,
   };
 
   #[test]
@@ -1015,9 +1021,13 @@ mod tests {
   #[test]
   fn user_class_named_buffer_precedes_the_builtin_buffer_mapping() {
     crate::typegen::r#struct::CLASS_STRUCTS.with(|classes| {
-      classes
-        .borrow_mut()
-        .insert("Buffer".to_owned(), "UserBuffer".to_owned());
+      classes.borrow_mut().insert(
+        "Buffer".to_owned(),
+        crate::typegen::r#struct::ClassStructRef {
+          js_name: "UserBuffer".to_owned(),
+          js_mod: None,
+        },
+      );
     });
     let ty = syn::parse_str("Buffer").expect("Buffer must parse as a Rust type");
     assert_eq!(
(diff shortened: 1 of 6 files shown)


Brooooooklyn:
@codex review

PR #3436, opened 2026-08-01
feat(cli): size WASI browser worker pools from navigator.hardwareConcurrency

Brooooooklyn:
## Why

Threaded browser loaders get two worker pools: the wasi-threads pool (`reuseWorker`, pre-created at init) and the emnapi async-work pool (`asyncWorkPoolSize`). Both were hardcoded to **4**, which undersizes both ends of the range:

- **Big machines**: on an 18-core desktop, a hash at `p=8` can't get its 7 helpers from a pool of 4.
- **Privacy-fuzzed browsers**: Tor/Brave report 2 cores, where a pool of 4 oversubscribes.

## What

Both pools are now sized at runtime from `navigator.hardwareConcurrency`:

```js
const __workerPoolSize = Math.max(
  2,
  (globalThis.navigator?.hardwareConcurrency ?? 4) - 1,
)
```

Logical cores minus one for the main thread, floored at 2, falling back to 4 when the API is missing or fuzzed away. Emitted for `asyncWorkPoolSize` and `reuseWorker` alike; threaded builds continue to initialize asynchronously (pool pre-creation is async).

Background on why the pool exists at all: a browser cannot start a worker while a thread is blocked inside a wasm call, so addon Rust code spawning threads mid-call deadlocks unless the pool is pre-created. This PR keeps that mechanism and just sizes it correctly.

## Measured

With `examples/argon2-bench` (Chromium, m=64 MiB, t=3, Argon2id, tags bit-identical):

| cell | before (pool of 4) | after (pool of 17) |
|---|---:|---:|
| p=1 | 68.7 ms | 68.7 ms |
| p=4 | 19.0 ms | 19.0 ms |
| p=8 | *(deadlock — pool exhausted)* | **11.9 ms** (5.8x vs p=1) |

## Tests

- `createWasiBrowserBinding threaded builds size worker pools from hardwareConcurrency`
- `createWasiBrowserBinding threadless keeps sync init and no pool`

**25/25 template tests pass**, snapshots regenerated. Backward compatible: threadless output is unchanged; threaded output only replaces the two constants with the computed binding.

---

> [!NOTE]
> **Medium Risk**
> Changes generated browser threaded-loader init and worker lifecycle (more workers on large machines, async-only path for threads); threadless bindings are untouched but mis-sized pools could still affect concurrency-heavy addons in browsers.
> 
> **Overview**
> Threaded **browser** WASI bindings no longer use a fixed pool of four workers. Generated loaders compute `__workerPoolSize` at runtime with `Math.max(2, navigator?.hardwareConcurrency ?? 4)` and pass `reuseWorker: { size: __asyncWorkPoolSize + __workerPoolSize }` so the pre-created pool scales with logical cores while reserving four slots for emnapi async work on the same reuse pool.
> 
> Because worker pool pre-creation is async, threaded builds always use **`instantiateNapiModule`** (even when `asyncInit` is false); threadless builds stay on sync init with `asyncWorkPoolSize: 0` and no `reuseWorker`. **`cli/docs/wasi.md`** documents the sizing formula and deadlock-avoidance rationale. Template tests and snapshots were updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6143cfd0dda0abce12bf0c441c1b0c077d460d7b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Improved threaded browser WASI loading by pre-creating and reusing workers.
  * Worker pool sizes now adapt to available browser hardware, with a minimum of two workers and a safe fallback.
  * Threaded browser modules initialize asynchronously for more reliable blocking operations.
  * Threadless builds retain synchronous initialization.
  * Worker pools can allocate additional workers when capacity is temporarily exhausted, avoiding blocking-operation deadlocks.


Brooooooklyn:
Thanks for the review, @coderabbitai. Both items were verified empirically before acting — one accepted with a composition fix, one's premise turned out partially incorrect.

**`strict: true` — accepted, with a critical correction.** Tracing the control flow through `@emnapi/wasi-threads`: `strict` makes exhaustion return `undefined` → `threadSpawn` writes `EAGAIN` into the result struct → Rust sees a recoverable spawn error. Two caveats the review missed:

1. **`strict` is browser-only** — emnapi's `!ENVIRONMENT_IS_NODE` guard means Node still silently allocates overflow workers, so "rejects spawning" isn't portable.
2. **Applying it naively regressed everything to single-thread speed**: emnapi's async-work pool draws from the *same* reuse pool, and sizing both pools from `hardwareConcurrency` left zero slots for addon thread spawns (measured: 492 "thread pool is exhausted" errors on the first hash, helpers all `EAGAIN`). `asyncWorkPoolSize` therefore stays a small constant (4), leaving `cores-5` slots for Rust threads.

Verified end-to-end in Chromium (`examples/argon2-bench`, m=64 MiB, t=3): p=4 19.0 ms, p=8 11.4 ms, p=32 9.3 ms (13 pooled helpers + graceful EAGAIN fallback for the rest), tags bit-identical.

**Test strengthening — applied, but the premise was partially off.** The review claimed `hardwareConcurrency` "also appears in the generated explanatory comment", making the assertion weak. It does not: the prose comment lives in the template *source*, not the emitted binding — the generated file contains that text only in the executable expression. I confirmed this by removing the expression (the test then fails), so the original assertion did guard the executable path. The stricter assertions (`const __workerPoolSize = Math.max(`, `strict: true`, threadless negative checks + `asyncWorkPoolSize: 0`) are applied anyway in 4da0c877.

Brooooooklyn:
Thanks for the P1 catch — verified empirically, you're right about the reservation, and it's fixed in `6143cfd0`. One part I have to push back on after testing.

**Accepted: the async-work reservation.** Traced it in Chromium: with the pool smaller than `initWorkers(4)`, async-work setup fails with `EAGAIN` — this was also the root cause of the "Test node wasi target" CI failure (`failed to get new worker` / WouldBlock in `readFileAsync` and `testWorkers`). The reuse pool is now `__asyncWorkPoolSize + hardwareConcurrency`, so async work always initializes and addon threads get the full core-derived capacity.

**Not applied: `strict`.** I implemented it and tested end-to-end — it breaks spawn-and-return workloads at exhaustion. The example's own `testWorkers` (4 workers + 1 joiner thread) hits the pool ceiling and its `std::thread::spawn` *panics* on `EAGAIN` ("Resource temporarily unavailable" → `RuntimeError: unreachable`). That's strictly worse than the non-strict fallback: the overflow spawn allocates a fresh worker which boots as soon as the spawning parent returns to its event loop — the mechanism by which both patterns complete (pooled for speed, fresh for overflow). Meanwhile the deadlock `strict` is meant to prevent only exists for joins inside a blocked call, and those draw from the pre-created pool by construction. So: reservation included, `strict` left out.

Verified locally with the CI condition simulated (`__workerPoolSize = 4`, matching a 4-vCPU runner): all 6 browser tests in `examples/napi` pass, including `readFileAsync` and `testWorkers`.

Issue #2573, opened 2025-04-23
NPM Package: option for all-in-one and seperated packages


Dexus:
Hi, i would like to create a all-in-one package, but it looks like this is not direct supported because its always build npm packages, for each os/arch.

I would like to use the same approche as prebuildify I build the binaries and ship them in the main package. 

Can this added?

Issue #1164, opened 2022-05-06
Class inheritance support


devongovett:
For Parcel CSS, I'm implementing the [CSS OM](https://drafts.csswg.org/cssom-1/) API, which includes many classes which inherit from one another. For example, `CSSRule <- CSSGroupingRule <- CSSConditionRule <- CSSMediaRule`. So far, I've been doing this by patching the prototype chain of the generated classes from the JavaScript side, along with some tricks on the Rust side to make the memory layout work.

On the Rust side:

```rust
#[napi]
#[repr(C)]
struct CSSRule {
  // ...
}

#[napi]
#[repr(C)]
struct CSSGroupingRule {
  parent: CSSRule,
  // ...
}
```

and the JS:

```js
Object.setPrototypeOf(CSSGroupingRule.prototype, CSSRule.prototype);
Object.setPrototypeOf(CSSGroupingRule, CSSRule);
```

Basically, the super class is always listed first in the subclass struct, and `#[repr(C)]` ensures that the fields won't get reordered by the Rust compiler. This means that when accessing a property or method from a super class in JS, it will be resolved in the prototype chain, and on the Rust side, the pointer will be viewed as the parent struct (fields after in the child will be ignored).

This works but it might be nice for napi-rs to do this for me. For example, you could write this instead:

```rust
#[napi(extends = CSSRule)]
struct CSSGroupingRule {
  // ...
}
```

and napi-rs could generate the above structure, and calls to setup the prototype chain properly.

## Problems with super class methods

One problem with this approach currently is that methods don't work when defined in a super class but accessed from an instance of a subclass. For example:

```rust
#[napi]
impl CSSRule {
  #[napi]
  pub fn super_method(&self) {}
}
```

```javascript
let rule = new CSSGroupingRule();
rule.superMethod();
```

This will fail with an "Illegal invocation" error from v8. I traced this back to a few issues in Node, e.g. https://github.com/nodejs/node-addon-api/issues/246, https://github.com/nodejs/node/issues/38038, https://github.com/nodejs/node/pull/20267, https://github.com/nodejs/node-addon-api/issues/229, etc.

The crux of it is this line in `napi_define_class`: https://github.com/nodejs/node/blame/55079bbebf02d12e56b7068ff0d3e32f9cf83bcc/src/js_native_api_v8.cc#L872-L873

[...]

Issue #1112, opened 2022-04-01
Intentionally support napi across multiple crates


Xaeroxe:
Imagine the following scenario: I have two crates that expose napi modules, I'd like to share some definitions between them and I'm not fond of duplicating code (Don't Repeat Yourself). So I move the shared definitions into a third crate, and depend on it from both napi modules.

The scenario I described almost works today! By accident. Like most things that happen by accident the implementation isn't quite perfect. All invocations of the `#[napi]` procedural macro which are evaluated in the same invocation will output their typescript definitions to the [same temporary file](https://github.com/napi-rs/napi-rs/blob/548c07e2446d1eaaf7e0ef17487084c30ae4c437/cli/src/build.ts#L204). In a workspace, this means that file can get polluted with types from other crates. We can use this to our advantage, but it doesn't always work, especially because not all `#[napi]` uses are necessarily going to be evaluated when we run `napi build`. Only the ones that Rust deems to need a recompile will be evaluated. Additionally, that temporary file may contain definitions for crates which were in the workspace, but not declared as dependencies.

So, currently, napi across multiple crates in a workspace kind of works by accident. I'd like to see it work well on purpose.

Xaeroxe:
https://doc.rust-lang.org/cargo/guide/build-cache.html#dep-info-files might be useful for implementing this

Xaeroxe:
Pursuant to this, I wonder if it'd be better to split code generation into two steps. Phase 1, during compilation, rather than generating a TypeScript file, a database file of the types and functions is generated (maybe SQLite?). The database is populated by the procedural macros when they get evaluated. Then, phase 2, post compilation the database is read from and used to generate the needed TypeScript and JavaScript files.

Executing in a two step manner like this alleviates a lot of the pain that comes from the short lived nature of procedural macros. Additionally, the database is much easier to retain and modify than a TypeScript file.

Xaeroxe:
Another benefit to the above approach is that if, for some reason, it's easier to determine a value at runtime than compile time, the typescript file generator could load the compiled artifact and call functions on it.

Issue #1040, opened 2022-01-18
Rebuilding without changes breaks the build output - no exports


AlCalzone:
When rebuilding a project created with `napi-rs` that has no changes on the Rust side, the binaries can no longer be used from JS code. 
When building from a clean state (deleted `target` and the `index.*` files), the native module can be required and works.
If you now rebuild again without changing any Rust code, the build finishes fairly quickly, but the native module is now broken and exposes no `exports` at all. Accessing them will fail with a `TypeError`.

I've made a repro repo here which demonstrates the issue: https://github.com/AlCalzone/napi-rs-repro-build-issue

PR #3437, opened 2026-08-03
feat: #[napi(extends)] inheritance + cross-crate identity, reflection, registration safety, and multi-binary builds

Boose101:
## Summary

This branch bundles a set of related napi-rs improvements, layered on top of
single-class inheritance for `#[napi]` structs. Opened as a **draft** for review.
The work is a stack of self-contained commit groups, each independently
reviewable and independently tested.

Everything is additive and backward compatible: existing single-crate,
single-binary projects compile and generate identical output (the type-def cache
identity is unchanged when no binary name is involved).

### Related issues

- `#[napi(extends)]` inheritance — implements a first version of
  napi-rs/napi-rs#1164
- Cross-crate type identity — relates to napi-rs/napi-rs#1112
- Multi-binary builds — relates to napi-rs/napi-rs#2573, and to
  napi-rs/napi-rs#1040 (the "rebuild drops exports" re-emission path this work
  depends on and exercises)

---

## `#[napi(extends = Parent)]` single-class inheritance  (napi-rs/napi-rs#1164)

**Model.** A child struct embeds its parent as the first field and is `#[repr(C)]`,
so a `*const Child` is a valid `*const Parent` at offset 0:

```rust
#[napi]
#[repr(C)]
pub struct Sub {
  parent: Base, // first field, no getter
  extra: i32,
}

#[napi(extends = Base)]
impl Sub { /* ... */ }
```

**Runtime (`crates/napi/src/bindgen_runtime/module_register.rs`).** Registration
computes a per-class tag and a parent tag, `build_hierarchy` validates the chain
(fail-closed on a cycle/missing parent, thrown as a JS error), and
`wire_extends_prototypes` links the **instance** prototype chain so `instanceof`
and inherited getters/setters/methods resolve. Receiver unwrap casts through the
offset-0 embedded parent via the type-tag check in `type_tag.rs`.

**v1 boundaries (intentional, documented on the macro rustdoc):** the
constructor-side prototype is not wired, so statics and factories are not
inherited (a child uses a factory, not a constructor, alongside `extends`);
inherited *plain methods* additionally require `napi8`; iterator combination is
out of scope.

**v1 hardening.** The narrow model is made harder to misuse and harder to break:

- **Clearer misuse diagnostics.** The generated first-field type check
  (`__NapiExactType` marker trait) now carries `#[diagnostic::on_unimplemented]`,
  so embedding the parent behind `Box`/`Rc`/a newtype — or putting it anywhere
  but first — produces a message that names the misuse and points at the field,
  instead of a raw unsatisfied-bound error.
- **Offset-0 invariant is asserted, not just assumed.** The unchecked runtime
  upcast reinterprets `*mut Child` as `*mut Parent` with no offset adjustment.
  `#[repr(C)]` + parent-first-field guarantee offset 0; a generated
  `const { offset_of!(Child, parent) == 0 }` (in its own scope, so it runs even
  when the type check fails) now makes that a self-checking invariant a future
  codegen/layout change cannot silently violate. The runtime cast site
  cross-references the guarantee.
- **Expanded compile-fail coverage.** Fixtures for a wrong/non-first first field,
  `extends` + `#[napi(constructor)]`, and an empty struct, alongside the existing
  missing-`repr(C)`, `repr(packed)`, non-class, and iterator cases.
- **Docs frame the scope.** The rustdoc states the two compile-time assertions,
  notes single-parent (multi-level chains supported), and names the planned
  opt-in `#[napi(extends = Parent, via = AsRef)]` composition variant as the
  growth path that would lift the `#[repr(C)]`/first-field constraint.

**Typegen** uses TypeScript declaration merging rather than `class extends`, so
the emitted types describe instance inheritance without claiming the static
inheritance the runtime omits:

```ts
export declare class Sub {
  static create(/* ... */): Sub
  extraKind(): number
}
export interface Sub extends Base {}
```

This is emitted as a distinct instance-extends type-def rendered as a sibling
`export interface X extends Y {}` (via the cross-crate-resolved parent name
below), not by threading `extends` into the class's own declaration.

## Canonical `(namespace, js_name)` type identity, provable across crates  (napi-rs/napi-rs#1112)

**Problem.** Backend type-name resolution keyed on the Rust struct ident and
produced a bare (namespace-less) js_name. Within one crate this cannot qualify a
cross-namespace reference; across crates a consumer's macro expansion never sees
the producer crate's `CLASS_STRUCTS`/`ALIAS` thread-locals at all (they are
process-local to each crate's proc-macro run), so the reference leaked out as the
bare Rust ident.

**Mechanism.** Because a proc-macro cannot resolve another crate's identity at
expansion time, resolution is deferred to the one stage that sees every crate's
fragments — the CLI join — reusing the existing marker/sideband pattern already
used for buffer imports:

- `crates/backend/src/typegen/struct.rs`: `CLASS_STRUCTS`/`ALIAS` now carry
  `(Option<js_mod>, js_name)`; each fragment records its producing crate via
  `env!("CARGO_PKG_NAME")`.
- `crates/backend/src/typegen.rs`: when a same-crate lookup misses,
  `handle_type_path`/`handle_generic_type` emit an unresolved-class **sentinel**
  (analogous to `BUFFER_TYPE_IMPORT_SENTINEL`) carrying the raw ident instead of
  falling back to it directly.
- `cli/src/utils/typegen.ts` (`preprocessTypeDef`): builds an index
  `original_name → [(namespace, js_name, producer_crate)]` across all fragments
  and resolves each sentinel — exactly one candidate → qualified reference
  (`Ns.Name` or bare at top level); zero → bare ident (a legitimate
  hand-written external type); more than one → **fail generation** naming every
  conflicting `(producer_crate, namespace)` pair rather than guessing.

**Alias policy.** The dead `module.exports.<RustIdent>` runtime alias (it pointed
at a non-existent export) is removed unconditionally. The
`export type <RustIdent> = <js_name>` TypeScript alias is preserved behind a new
`emitRustNameTypeAlias` config flag (**default `true`**); a same-ident collision
across namespaces makes the alias emission a hard generation error.

## Safe reflection helpers on JS values

Three new helpers, each with sharpened, documented, and tested semantics — and
each **compiled out** (rather than silently returning a wrong answer) outside its
supported feature/platform:

- `Object::to_serde_json_value(&self) -> Result<serde_json::Value>` (gated on
  `serde-json` + `napi6`), with documented behavior for `undefined`, functions,
  symbols, `BigInt`, circular refs, `Date`, typed arrays, class instances,
  throwing getters, and Proxy traps.
- `constructor_name(&self) -> Result<Option<String>>` — `Ok(None)` for
  null-prototype objects or a missing/non-string `constructor.name`, `Err` only
  when the property access itself throws (documented: reading `constructor` can
  run JS getters/Proxy traps).
- `has_type_tag::<T: TypeTag>(&self) -> Result<bool>`
  (`cfg(all(napi8, not(wasm)))`), so "the check ran and said no" can never be
  confused with "this platform can't perform the check".

## Fail-closed duplicate-registration guard

**Problem.** N-API registration defines exports with `napi_set_named_property` /
`napi_define_class`, both of which silently let a later definition overwrite an
earlier same-named one — a duplicated export or member just disappears at runtime
with no error.

**Mechanism.** A pre-pass in `napi_register_module_v1`
(`crates/napi/src/bindgen_runtime/module_register.rs`), run before the drain
loops, walks `MODULE_REGISTER_CALLBACK` (functions) and `MODULE_CLASS_PROPERTIES`
(classes, already merged per class) into a descriptor list and calls the pure
`detect_registration_conflicts`, which fails closed (same `Err(String)` → throw
path as `build_hierarchy`) on a genuine collision. It is **descriptor-level**, so
valid patterns are not flagged:

- module scope: a class and a function sharing `(js_mod, js_name)` → error;
- within a class: getter + setter for one name → OK; getter + method, two
  methods, two getters, or two setters for one name → error; a static and an
  instance member of one name → OK; a class assembled from several
  `#[napi] impl` blocks → OK (the existing merge path is not regressed).

Ordering is deterministic (`BTreeMap`), so the reported conflict does not depend
on linker/constructor execution order. Exposed Rust-only — no new JS surface.

## Multi-binary builds (`binaries[]`)  (napi-rs/napi-rs#2573, napi-rs/napi-rs#1040)

**Config.** A `binaries[]` array builds more than one output addon from a single
`napi build`:

```jsonc
"napi": {
  "binaries": [
    { "name": "full",   "manifestPath": "crates/full/Cargo.toml",   "binaryName": "index",        "js": "index.js",        "dts": "index.d.ts" },
    { "name": "client", "manifestPath": "crates/client/Cargo.toml", "binaryName": "index-client", "js": "index-client.js", "dts": "index-client.d.ts" }
  ]
}
```

`name` is a stable selector (independent of the output `binaryName`); each entry
takes exactly one of `manifestPath`/`package`.

**Pipeline (`cli/src/api/build.ts`).** `buildProject` reads the config once, then
a pure, unit-tested `normalizeConfig(config, options)` expands it:

- no `binaries[]` → a single spec, the classic build unchanged;
- `binaries[]` → one spec per entry (or the `--binary <name>` selection), after
  validating unique `name`, exactly-one-of `manifestPath`/`package`, no nested
  `binaries`, and no two entries writing the same `.node`/`js`/`dts`.

Each spec runs through `buildSingleProject`; `buildProject` loops them,
concatenating outputs and preserving its `{ task, abort }` return contract.
`--watch` with `binaries[]` is rejected.

**Cache isolation.** `getTypeDefCacheFolder` gains an optional binary `name` in
its identity, so two binaries built from the same crate get separate type-def
caches. The name is omitted for single-binary builds, so their existing cache
folder hash is unchanged.

> **Re-emission (relates to #1040).** For a shared dependency crate's type-defs
> to appear in every consuming binary, the shared crate must call
> `napi_build::setup()` in its own `build.rs`: that build script emits
> `cargo::rerun-if-env-changed=NAPI_FORCE_BUILD_<CRATE>`, which lets the CLI
> force the (otherwise fully cached) shared crate to recompile and re-emit its
> fragment into each binary's own cache folder. With it, per-binary caches stay
> correct from a fully warm build — no separate `--target-dir` needed. A crate
> without a `build.rs` never re-emits once cached, which is the same root cause
> behind #1040.

---

## Testing

- **Rust** — `cargo test --workspace`: inheritance suite (accessors, multi-level
  chains, exact-vs-descendant receiver validation, GC finalize/Drop via native
  atomic counters, concurrent-worker fail-fast on a broken hierarchy, no-`napi8`
  inherited-method guard), reflection helpers, and the descriptor-level
  duplicate-registration unit tests (every rule above).
- **Compile-fail** — trybuild cases for invalid inheritance layouts: missing
  `#[repr(C)]`, `#[repr(packed)]`, non-class target, iterator combination,
  `extends` + `#[napi(constructor)]`, empty struct, and a first field that is
  not exactly the parent (asserting the `on_unimplemented` message).
- **TypeScript** — ava specs and typegen snapshots, incl. a positive/negative
  `tsc` check proving instance-only inheritance (a child's static access fails to
  compile) and a cross-crate identity fixture asserting qualified references.
- **CLI** — pure unit tests for `normalizeConfig` (expansion, every validation
  rule, `--binary` selection, per-binary cache keying), plus a real-cargo
  warm-cache matrix over a two-binary workspace: clean and reverse-order builds
  produce identical type defs; a fully warm compile still regenerates deleted
  `js`/`dts`; selecting one binary leaves the other's output untouched (both
  directions); the slim client never leaks a full-only export; the shared
  export renders identically in both; and a consumer of each generated `.d.ts`
  type-checks (via the TypeScript compiler API).

## Status

Draft. The commit groups are self-contained and each is independently
reviewable and tested; happy to split them into separate stacked PRs if that
is easier to review.

## Summary by CodeRabbit

* **New Features**
  * Added single- and multi-level N-API class inheritance with inherited methods, accessors, prototype wiring, and runtime type validation.
  * Added multi-binary builds with per-binary configuration, isolated outputs, `--binary` selection, and independent type-definition caches.
  * Added reflection helpers for constructor names, type tags, and JSON conversion.
  * Added cross-package type references and optional suppression of Rust-name TypeScript aliases.

* **Bug Fixes**
  * Improved validation and diagnostics for duplicate exports, invalid hierarchies, and unsupported inheritance layouts.
  * Added safeguards against excessively deep or cyclic JavaScript-to-value conversions.
  * Improved inherited object cleanup and receiver handling across supported runtime configurations.


Boose101:
@Brooooooklyn What are your thoughts on this?

Boose101:
@coderabbitai review

Boose101:
@coderabbitai resume

Boose101:
@Brooooooklyn Would you prefer if i split this into multiple PR's for ingestion / readability? Would love to get these changes landed and think they would resolve some of the issues that have been long existing and add new functionality that is needed. Let me know how to best make this ingestible / reviewable for you. Thank you so much!

PR #3464, opened 2026-08-24
perf(napi): outline class instance construction

Boshen:
Share class instance construction and reference-map setup across class types.

Rolldown measurement: `__text` 11,962,304 → 11,938,740 bytes (-23,564 B); stripped dylib 16,506,176 → 16,489,608 bytes (-16,568 B).

## Summary by CodeRabbit

* **Refactor**
  * Improved internal object creation and reference handling.
  * Preserved existing instance creation behavior and public APIs.
  * No user-visible functionality changes.


Brooooooklyn:
@codex review

PR #3466, opened 2026-08-24
perf(sys): deduplicate dynamic symbol loading

Boshen:
Load Node-API exports through one out-of-line raw symbol lookup before converting each address to its declared ABI signature.

Rolldown fixed-path measurement: `__text` 11,962,304 → 11,933,576 bytes (-28,728 B); stripped dylib 16,506,176 → 16,489,624 bytes (-16,552 B).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility when loading platform-specific Node-API symbols.
  * Preserved fallback behavior when optional symbols are unavailable.

### DIFF
--- a/crates/sys/src/lib.rs
+++ b/crates/sys/src/lib.rs
@@ -6,6 +6,21 @@
   target_env = "msvc",
   all(not(target_family = "wasm"), feature = "dyn-symbols")
 ))]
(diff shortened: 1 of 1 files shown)


Brooooooklyn:
@codex review

PR #3477, opened 2026-09-02
chore(napi): keep anyhow error cause when converting

kasper9n:
This makes the `cause` be added to napi::Error and napi::JsError when converting from anyhow errors.

Napi currently has an inconsistency where one of the conversions includes the causes in the error message, while the other does not. With this PR, causes only appear in the `cause` property, not in the `message`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages by preserving the full chain of underlying causes.
  * JavaScript errors now expose structured cause information instead of combining all details into a single text message.


Brooooooklyn:
@codex review

PR #3486, opened 2026-09-07
fix(macro): recognize fully-qualified napi::Env as the special Env parameter

### LINKED ISSUE #1597: Cannot use `#[napi]` on function when parameter of type `napi::Env` is present
Using the qualified names `napi::Env` or `napi::env::Env` for arguments causes the `napi` macro to fail, whereas importing `napi::Env` and using the imported `Env` as type works.

Diagnostic:
```
error[E0277]: the trait bound `Env: NapiValue` is not satisfied
   --> src\lib.rs:357:1
    |
357 | #[napi]
    | ^^^^^^^ the trait `NapiValue` is not implemented for `Env`
    |
    = help: the following other types implement trait `NapiValue`:
              Date
              JsArrayBuffer
              JsBigInt
              JsBoolean
              JsBuffer
              JsDataView
              JsExternal
              JsFunction
            and 10 others
    = note: required for `Env` to implement `FromNapiValue`
    = note: this error originates in the attribute macro `napi` (in Nightly builds, run with -Z macro-backtrace for more info)
```

SamyakJ05:
## Summary

Fixes #1597.

`#[napi]` fails to compile when a function parameter uses the fully-qualified path `napi::Env` (or `napi::env::Env`) instead of the bare `Env` identifier, with:

```
error[E0277]: the trait bound `Env: NapiValue` is not satisfied
```

## Root cause

The macro special-cases an `Env`-typed parameter so it receives the callback's `Env` directly instead of going through the normal `FromNapiValue` conversion path. In `crates/backend/src/codegen/fn.rs` this check was:

```rust
if &pat_type.ty.to_token_stream().to_string() == "Env" {
```

This compares the *entire* type's token stream against the literal string `"Env"`. That only matches the bare identifier — a qualified path like `napi::Env` renders as `"napi :: Env"`, which never equals `"Env"`, so the special case silently fails to trigger and the parameter falls through to ordinary `FromNapiValue`-based extraction (which doesn't exist for `Env`), producing the confusing trait-bound error from the issue.

The same whole-token-stream comparison pattern also existed in `crates/backend/src/typegen/fn.rs` (used for `.d.ts` generation), with the same defect.

## Fix

Both places now match on the resolved type's **last path segment ident** instead of the whole token stream — the same idiom already used elsewhere in these exact files for other special-cased parameter types (`Reference`, `This`, `WeakReference`) and in `crates/macro/src/parser/mod.rs` for `Env`/`Object` detection in `module_exports`. This makes recognition independent of how many path segments precede `Env`, so bare `Env`, `napi::Env`, and `napi::env::Env` are all now handled consistently.

All Env checks also require `qself: None`, so a qualified-self path like `<T as Trait>::Env` is never treated as the special environment parameter. That guard covers the borrowed-parameter branch in `codegen/fn.rs` too — without it, `&<T as Trait>::Env` was injected as the environment in codegen while typegen (already guarded) kept it as a regular argument, making the generated `.d.ts` disagree with runtime argument consumption.

## Test plan

- Added `with_fully_qualified_env` to `examples/napi-cargo-test/src/lib.rs`: a `#[napi]` function taking `napi::Env` (fully qualified) as a parameter, which only compiles successfully with this fix (verified it fails with the pre-fix code, reproducing the exact error from the issue).
- Added `with_qualified_self_env_associated_type`: a `#[napi]` function taking `<MyEnvHolder as HasEnv>::Env` (a qualified-self associated type resolving to `i32`), which only compiles when the parameter is extracted normally instead of being special-cased as the environment.
- Added `with_qualified_self_env_ref`: the borrowed form, `&<MyEnvRefHolder as HasEnvRef>::Env` resolving to a `#[napi]` class, which only compiles when it goes through normal `FromNapiRef` extraction (verified it fails with E0308 without the borrowed-branch guard).
- `cargo test -p napi-cargo-test` — 7 passed
- `cargo test -p napi-derive-backend -p napi-derive` — 17 passed (backend unit tests + macro crate)
- `cargo test -p napi` — 10 passed
- `cargo check -p napi-examples` (the full `examples/napi` crate, which exercises bare `Env` and `&Env` parameters extensively) — compiles clean, no regressions

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved generated TypeScript signatures by correctly distinguishing the special environment parameter from qualified-self associated types named `Env`.
  - Qualified-self `Env` associated types are now retained as regular function arguments and processed through standard value conversion, including borrowed values.

- **Tests**
  - Added regression coverage for qualified-self associated types named `Env`, including owned and borrowed forms, extraction, and runtime behavior.

### DIFF
--- a/crates/backend/src/codegen/fn.rs
+++ b/crates/backend/src/codegen/fn.rs
@@ -679,8 +679,10 @@ impl NapiFn {
             }
           }
         } else {
-          if let syn::Type::Path(ele) = &*elem {
-            if let Some(syn::PathSegment { ident, .. }) = ele.path.segments.last() {
+          // `qself: None` keeps qualified-self paths like `&<T as Trait>::Env`
+          // on the normal `FromNapiRef` extraction path, matching typegen.
+          if let syn::Type::Path(syn::TypePath { qself: None, path }) = &*elem {
+            if let Some(syn::PathSegment { ident, .. }) = path.segments.last() {
               if ident == "Env" {
                 return Ok((quote! {}, NapiArgType::Env));
               } else if ident == "str" {
--- a/crates/backend/src/codegen/fn.rs
+++ b/crates/backend/src/codegen/fn.rs
@@ -449,7 +449,16 @@ impl NapiFn {
 
       match &arg.kind {
         NapiFnArgKind::PatType(pat_type) => {
-          if &pat_type.ty.to_token_stream().to_string() == "Env" {
+          let is_env_type = if let syn::Type::Path(syn::TypePath {
+            path: syn::Path { segments, .. },
+            ..
+          }) = pat_type.ty.as_ref()
+          {
+            segments.last().is_some_and(|s| s.ident == "Env")
+          } else {
+            false
+          };
+          if is_env_type {
             args.push(quote! { __wrapped_env });
             skipped_arg_count += 1;
           } else {
--- a/crates/backend/src/codegen/fn.rs
+++ b/crates/backend/src/codegen/fn.rs
@@ -450,8 +450,8 @@ impl NapiFn {
       match &arg.kind {
         NapiFnArgKind::PatType(pat_type) => {
           let is_env_type = if let syn::Type::Path(syn::TypePath {
+            qself: None,
             path: syn::Path { segments, .. },
-            ..
           }) = pat_type.ty.as_ref()
           {
             segments.last().is_some_and(|s| s.ident == "Env")
--- a/crates/backend/src/typegen/fn.rs
+++ b/crates/backend/src/typegen/fn.rs
@@ -197,9 +197,14 @@ impl NapiFn {
         .iter()
         .filter_map(|arg| match &arg.kind {
           crate::NapiFnArgKind::PatType(path) => {
-            let ty_string = path.ty.to_token_stream().to_string();
-            if ty_string == "Env" {
-              return None;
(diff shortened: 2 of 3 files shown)


Brooooooklyn:
Maintainer follow-up (19971fe1): applied the missing `qself: None` guard to the **borrowed** Env check in `codegen/fn.rs`. The earlier commits guarded the by-value checks and the typegen reference check, but not the codegen reference branch — so `&<T as Trait>::Env` was still injected as the environment in codegen while typegen kept it as a regular argument, and the generated `.d.ts` disagreed with runtime argument consumption. This also adds the referenced-form regression test (`with_qualified_self_env_ref`, using a `#[napi]` class as the associated type), which was the remaining item from the earlier review; verified it fails with E0308 without the guard.

One trade-off to be aware of (not changed by this follow-up): Env detection is name-based on the last path segment, the same idiom this codebase already uses for `Reference`/`This`/`WeakReference` and `module_exports` validation. That means any plain path ending in `Env` (e.g. a user's own `config::Env`) is now treated as the napi environment — it fails loudly with a type mismatch at compile time rather than silently. Restricting recognition to `Env`/`napi::…` spellings instead would avoid that but break re-exported paths like `my_crate::Env`. Went with codebase consistency here; happy to switch if you'd prefer the stricter matching.

Brooooooklyn:
@codex review

Brooooooklyn (approved):

PR #3513, opened 2026-09-14
fix(cli): normalize NAPI_RS_ASYNC_WORK_POOL_SIZE in the Node WASI loader

Brooooooklyn:
## What

The threaded Node WASI loader now normalizes `NAPI_RS_ASYNC_WORK_POOL_SIZE`
(and its `UV_THREADPOOL_SIZE` fallback) before handing it to emnapi. The
generated loader gains a module-scope helper:

```js
function __normalizeAsyncWorkPoolSize(value) {
  const numeric = Number(value)
  if (!Number.isFinite(numeric)) {
    return 4
  }
  const integer = Math.trunc(numeric)
  return integer > 0 && integer <= 1024 ? integer : 4
}

const __asyncWorkPoolSize = __normalizeAsyncWorkPoolSize(
  process.env.NAPI_RS_ASYNC_WORK_POOL_SIZE ?? process.env.UV_THREADPOOL_SIZE,
)
```

The environment is read once, at module scope, and the option is passed as
`asyncWorkPoolSize: __asyncWorkPoolSize` — the identifier the browser loader
already uses. Threadless Node, browser and `./workerd` loaders are untouched:
they pass a literal `0` or a constant and never read `process.env`.

## Why

emnapi does not use `asyncWorkPoolSize` as given. `createNapiModule` in
`@emnapi/core` coerces it with `>> 0` (ToInt32), clamps the result into
`[-1024, 1024]`, and exposes `Math.abs(...)` of it to the wasm module as the
`_emnapi_async_work_pool_size()` import. Every threaded build links the
`emnapi-napi-rs-mt` archive (`crates/build/src/wasi.rs`), whose **in-wasm libuv
threadpool** sizes itself from that import:

```
size = _emnapi_async_work_pool_size();
if (size > 0) nthreads = size;
else          nthreads = UV_THREADPOOL_SIZE from getenv, else 4;
if (nthreads > MAX_THREADPOOL_SIZE /* 1024 */) nthreads = 1024;
for (i = 0; i < nthreads; i++)
  if (uv_thread_create_ex(...)) abort();
```

So the old bare `> 0` guard was wrong in **both** directions:

| env value | ToInt32 | emnapi reports | in-wasm pool | outcome before |
|---|---|---|---|---|
| `4294967296` (and any multiple of 2^32) | `0` | `0` | "unset" → `UV_THREADPOOL_SIZE`, else 4 | the requested size vanishes silently |
| `2147483648` | `-2147483648` | clamped `-1024` → `1024` | 1024 threads | `abort()` → `RuntimeError: unreachable` |

The pool is built lazily, on the first async-work submission, and one failed
thread create aborts the whole module. The fix maps everything outside
`[1, 1024]` to the default `4` — falling back rather than clamping — so emnapi
is never handed a number ToInt32 would change, and an unreachable request never
turns into a *different* pool size the caller did not ask for.

## Verification

Real threaded example loader, real Workers, the same `example.wasm32-wasi.wasm`
on both sides; one `withoutAbortController(1, 2)` call per run.
`base` = `origin/main`, `head` = this branch.

| `NAPI_RS_ASYNC_WORK_POOL_SIZE` | base | head |
|---|---|---|
| unset / `4` / `abc` / `Infinity` | ok, 4 threads | ok, 4 threads |
| `64` | ok, 64 threads | ok, 64 threads |
| `256` | ok, 256 threads | ok, 256 threads |
| `384` | trap | trap (in range, does not fit this build) |
| `512` | trap | trap (in range, does not fit wasm32) |
| `1024` | trap | trap (in range, does not fit wasm32) |
| `1025` | trap (emnapi substitutes 1024) | **ok, 4 threads** |
| `2048` | trap (emnapi substitutes 1024) | **ok, 4 threads** |
| `2147483648` | trap (ToInt32 negative → 1024) | **ok, 4 threads** |
| `4294967296` | ok, 4 threads — the setting silently vanished | ok, 4 threads, by the loader's own rule |

Thread counts are read back from linear-memory growth across the first async
call (8 MiB per pool thread). The rows that still trap are in `[1, 1024]` and
so are passed through unchanged by design — they are genuine "this pool does
not fit in wasm32" failures, identical on `main`, not something this PR can or
should paper over. What changes is only the out-of-range rows.

"trap" is `RuntimeError: unreachable` out of libuv's `abort()`. Each pool thread
reserves 8 MiB of linear memory (wasi-libc's `DEFAULT_STACK_MAX`), so this
example build — 16,384 initial pages — runs out at 384 threads; `2^32 / 8 MiB`
puts the hard wasm32 ceiling at 512 even with minimal initial memory.

- `yarn test:cli` — 231 passed (was 229; two new template cases: `1024` kept,
  `1025` defaulted).
- `examples/napi/__tests__/wasi-async-work-pool-size.spec.ts` — new subprocess
  test on the threaded `test-node-wasi` lane (`WASI_TEST=true`). It runs the
  real loader with `4294967296` and `2048` and asserts the async export
  resolves on a pool of four (32 MiB of linear-memory growth). Verified it
  fails on `origin/main` for `2048` and on the first draft of this change for
  both — it is a real guard, not a tautology.
- `yarn lint`, `yarn format:check` — clean.

## Compatibility

Behavior only changes for values that were already broken or unusable:

- `1`…`1024` — unchanged.
- `> 1024`, `<= 0`, fractional-below-one, non-numeric, `Infinity` — now `4`.
  Previously they either silently lost the setting or aborted the module.
- Threadless Node / browser / `./workerd` loaders — byte-identical output.

The generated loader text changed, so any downstream tooling that string-patches
the old `asyncWorkPoolSize: (function () { ... })()` IIFE anchor will fail
loudly rather than silently no-op.

## Ceiling

`1024` is kept because it is emnapi's own clamp and libuv's
`MAX_THREADPOOL_SIZE` — above it emnapi substitutes a value of its own, which
is exactly what this change avoids. It is *not* a promise that a 1,024-thread
pool can be allocated: with wasi-libc's default 8 MiB thread stacks a wasm32
module tops out near 500 threads, and whether a given size fits depends on
`napi.wasm.initialMemory` / `napi.wasm.maximumMemory`. Lowering the accepted
ceiling to something a wasm32 module can actually allocate is a separate
product decision, documented in `cli/docs/wasi.md` but deliberately not made
here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01T34fMrYCf2V13Mg8BKmnda


Brooooooklyn:
Closing: not needed. With the napi-rs threaded archive, emnapi's in-wasm libuv pool already treats an out-of-range value as unset and falls back to UV_THREADPOOL_SIZE or 4, so the loader does not need its own normalization. rolldown drops its matching loader patch instead.

